PR #4, opened 2017-08-07
Update api and build process

hbibkrim:
For https://github.com/CityOfZion/neon-wallet/issues/33

Updated API to be on par with neon-wallet.
Updated build process to package into node library.
Updated tests to test from packaged index.js.
Updated README.

Still can be polished further but good stop point to import into neon-wallet.


hbibkrim:
If there are any updates to neon-wallet/src/wallet, I will need to update this branch before merge

Ejhfast:
Looks great! Can we keep the quick start in the readme though?

Ejhfast:
Awesome work!

PR #5, opened 2017-08-08
Display GAS price

hbibkrim:
This is a breaking change, as I've renamed the getMarketPriceUSD method and changing the type of value it returns.

I've introduced the currency parameter so in the future, we could display the neo/gas value in different currencies!

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -11044,21 +11044,29 @@ var getBalance = exports.getBalance = function getBalance(net, address) {
 };
 
 // Returns the current asset price according to coinmarketcap
-var getAssetMarketPrice = exports.getAssetMarketPrice = function getAssetMarketPrice(amount, currency, asset) {
-  return _axios2.default.get('https://api.coinmarketcap.com/v1/ticker/' + asset + '/?convert=' + currency).then(function (response) {
-    var price = response.data[0]['price_' + currency];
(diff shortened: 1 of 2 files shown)


dop89:
Hi @hbibkrim please have a look at this PR: https://github.com/CityOfZion/neon-js/pull/7

I made a similar change as you did, but in my opinion the formatting of the currency should be separated of the fetching of the price. Your implementation does not consider locale specify formats, like the decimal symbol, the thousands separator and the placement of the currency symbol.  

In Germany, a valid currency string would look like this `1.234,56 €` whilst in the US `$1,234.56`.

hbibkrim:
@dop89 That is a good point. Yours is the better solution. Can you include an assetCode parameter as well so it'd be easier to iterate to retrieving the gas value? I probably shouldn't have removed the existing getMarketPriceUSD function either and deprecated it instead like you did.

dop89:
Ok, I'll do this. And I also copy your `getMarketPrices` function.

hbibkrim:
@dop89 Can you handle the formatting of the prices depending on the currencyCode and return the formatted price inside neon-js as well instead of neon-wallet?

hbibkrim:
What do you think? @canesin @Ejhfast

dop89:
@hbibkrim I updated my PR: https://github.com/CityOfZion/neon-js/pull/7

hbibkrim:
Closing PR as API used to get price has been reverted to bittrex from coinmarketcap as coinmarketcap was unreliable

PR #11, opened 2017-08-21
Ported to ts

rickyhan:


snowypowers:
Seperated into its own repo

PR #20, opened 2017-09-23
Replace constants api.TESTNET and api.MAINNET

0xfe:
It seems like these constants don't exist anymore -- replaced with strings `MainNet` and `TestNet`.


snowypowers:
Closing due to inactivity. This is also noted and `dev` has a `consts` file to keep track of all constants

PR #27, opened 2017-10-03
Smart Contract Script Builder

snowypowers:
## Feature

The Script Builder is basically a copy of[ NEO-VM's script builder](https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/ScriptBuilder.cs), slightly remodeled for JS (because we can't overload in JS 😞 )

This will introduce the capability to parse operations and params into a script that can be carried by InvocationTransaction to trigger smart contracts. There is a basic test suite that tests basic NEP5 methods. Scripts have been hand tested through `invokescript` on a self hosted v2.3.3 node to make sure they are working.

## Sidenotes

- Added an .editorconfig file that defaults to 2 spacing. I find this file more handy than commiting a .vscode folder for now because it is more general.

- Testing globals (chai) is now extracted into a `common.js` which is preloaded by the test script. This eliminates the crappy feeling of copying `import chai from 'chai'` for every new test file.


src/transactions/index.js
```diff
@@ -1,102 +1,102 @@
...
+  const ss = new StringStream(data)
+  let tx = {}
+  tx.type = parseInt(ss.read(1), 16)
+  tx.version = parseInt(ss.read(1), 16)
+  const exclusiveData = deserialize.exclusiveData[tx.type](ss)
+  tx.attributes = []
+  tx.inputs = []
+  tx.outputs = []
+  tx.scripts = []
+  const attrLength = ss.readVarInt()
```
localhuman: Due to the possibililty of DOS via TX Attributes, 
readVarInt should accept a `max` value, beyond which no more will be read.  ( see (https://github.com/neo-project/neo/blob/master/neo/Core/Transaction.cs#L194)

Same thing goes when writing/serializing a TX. You should not be able to add more than `MaxTransactionAttributes` ( which is currently 16)

snowypowers: The c# implementation ignores the rest of the attributes if more than 16. This is okay from the node viewpoint. However, as an interaction SDK, should we be throwing an error here instead?

I think that we should be warning the user that this will happen and not let it pass silently.

Satoshinaire:
This PR seems to have cleaned up a bunch of the `console.log` entries from `src/utils.js`, but those changes haven't been compiled into `lib/index.js`.

snowypowers:
Not gonna compile it until we stabilise it and do a proper release. There might be breaking changes that will break latest builds for Neon Wallet and we do not want that.

Satoshinaire:
@snowypowers Oohh... Neon pulls in from `master`? 😱

snowypowers:
OK. I just checked. Neon wallet pulls a specific commit so we wont affect it. However, i would like to only update `lib/index.js` when we are ready. Doing that means we should tag as release imo. Not sure about the flow here but its up to Ethan to release when he feels like it.

Satoshinaire:
Seems weird to me that there's committed code that doesn't reflect changes that have been made. If it was ready for merge, surely it was ready for merge in both places?

Just my opinion of course, and happy to defer to whatever Ethan decides.

I'm really just anxious to get rid of that dang `console.log` 😛

Ejhfast:
Yeah let's rebuild lib/index when it is fully updated

Ejhfast:
Will be pretty soon, log won't be around long 😊

PR #33, opened 2017-10-06
Transaction construction

Ejhfast:
Migrate to cleaner transaction construction across API (breaking changes)


Satoshinaire:
Nice! And we got a fresh `/lib/index.js` in there 😉

Ejhfast:
Just burned me... :p done in 5f8fc4e6bec51b396d6d81d0c8622d5f697519b2

snowypowers:
bump version since there is breaking changes?

Ejhfast:
After RPX is taken care of, will do that and other logistical things
(delete unused methods etc.)

snowypowers:
but will this commit cause others relying on your lib to break? Following your import method on your readme, they will be receiving the breaking changes without seeing a version bump, causing issues.

Ejhfast:
you can't index a package.json on version number right now, since no way to
do that from github (need to index on commit or maybe branch) and the
package is not on npm... i will do now just to be sure though, you are
right it is better policy

Satoshinaire:
Tagging versions should help, right? People developing ought to pin their import to a tag.

Ejhfast:
done! i never got indexing on tags to work, personally

PR #40, opened 2017-10-14
bugfix to _emitNum w/unit tests

birmas:
_emitNum was returning invalid results i.e 500 would return 1f4 whereas the bytecode should be f401

* _emitNum() reversed order of hex bytes and ensured leading 0 added to msb (when applicable)
* added unit tests for -1, 0, 13, 500, 65536
* utils.js: hex2bytes() convert a hex string to an array and optionally reverse order
* utils.js:  int2hex() convert an integer to hex and add leading zeros


Ejhfast:
thanks @birmas!

PR #56, opened 2017-11-05
serializeWitness should use varint script lengths

notatestuser:
Verified in the Neo source: https://git.io/vFWYR


snowypowers:
Good catch!

PR #57, opened 2017-11-05
Integration: Ledger support + Unencoded public key support to Dev

snowypowers:
This PR integrates the changes from `master` to `dev`:

- Add `signingFunction` to NeonDB's `doClaimAllGas` and `doSendAsset`. It is okay to add this functionality to NeonDB's API as fundamentally, the NeonDB API is meant to support neon-wallet and its needs.
- Add support to encode / decode public keys. We still do not differentiate between the two forms in names as I do not see a need yet. Most (all) of the code uses the encoded form. 
- Add support to Account class for getting the decoded public key through `getPublicKey(encode)`. 
- `isPublicKey` now supports an optional parameter to specify the encoding. If not provided, `isPublicKey` will return true for both forms.


Satoshinaire:
LGTM 😄

PR #59, opened 2017-11-07
Enable usage on private blockchain

ambethia:
Includes test. Allows for usage like:

```JavaScript
const PrivNet = 'http://localhost:5000'

neon.getBalance(PrivNet, 'AP9GA3EVUbfLNMSWiZLu8fDZQs388Z1kVU').then(data => {
  console.log(data)
})
```

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -51,7 +51,7 @@ export const allAssetIds = [neoId, gasId]
 
 /**
  * Perform a ClaimTransaction for all available GAS
- * @param {string} net - 'MainNet' or 'TestNet'.
+ * @param {string} net - 'MainNet', 'TestNet'.
  * @param {string} fromWif - WIF key of address you are claiming from.
  * @return {Promise<Response>} RPC response from sending transaction
  */
@@ -272,7 +272,7 @@ export const getClaimAmounts = (net, address) => {
  * @return {Promise<string>} The URL of the best performing node or the custom URL provided.
  */
 export const getRPCEndpoint = (net) => {
-  if (net !== 'TestNet' && net !== 'MainNet') return Promise.resolve(net)
+  // if (net !== 'TestNet' && net !== 'MainNet') return Promise.resolve(net)
   const apiEndpoint = getAPIEndpoint(net)
   return axios.get(apiEndpoint + '/v2/network/best_node').then((response) => {
     return response.data.node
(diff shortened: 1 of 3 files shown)


snowypowers:
Hi, looks like a pretty niche case but I think there can be use. Can this be rewritten and pointed at `dev` instead? We are trying to move forward with a new version. `master` will only accept urgent fixes.

You would be editing here : https://github.com/CityOfZion/neon-js/blob/cc571a23912cf349dfc7f55ae9dd806865e587c4/src/api/NeonDB.js#L53

ambethia:
Yes absolutely I can change the branch.

However I noticed later that there's a line in `getRPCEndpoint` later that breaks this, where that function does something similar. Commenting that line out works, because it will just use `neon-wallet-db` API to get the best RPC node.

snowypowers:
hmm don't they serve different purposes? `getAPIEndPoint` returns `neon-wallet-db` node while `getRPCEndPoint` returns a NEO RPC node. They have different use cases.

```js
// Here your change will work to allow a custom neon-wallet-db node.
// A neon-wallet-db is used to retrieve information. A custom neon-wallet node allows me to build my own light wallet and not rely on COZ nodes.
getBalance('http://myneonwalletnode:3000', addr)

// Here i want to use a custom NEO node, not a custom neon-wallet-db node.
// A NEO node is used here because i want to send a transaction to my own node. This is useful if I am a big company and I want to own the NEO node.
doClaimAllGas('http://myneonode:10332', privateKey)
```

Do correct me if im wrong.

notatestuser:
Would be awesome to see how you've set up neon-wallet-db to work with a private network if you have that code to drop in a PR! This is a great start.

ambethia:
@snowypowers Yes, they do return different endpoints, the problem is, currently there's a hack into `getRPCEndPoint` that does something similar to what my PR does, but only for the RPC endpoints, and it breaks with my change. I solved it by just deleting this line:

https://github.com/CityOfZion/neon-js/blob/master/src/api.js#L272

Unsure of how this PR should proceed, unless that change is also acceptable.

ambethia:
@notatestuser The changes I made are pretty hackey, and not suitable for a PR, but here's a diff to see what I changed.

https://gist.github.com/ambethia/7515ca068548d209429f5eed3b4b3623

I've found it to be pretty finicky, and I get tons of errors in the console, but it mostly works :D

ambethia:
Closed in favor of #62

PR #62, opened 2017-11-08
Enable usage on private blockchain

ambethia:
Replaces #59


snowypowers:
OK perfect. I thought about it and would say that this change would make more sense. 

The aim of NeonDB API is to support transactions built using `neon-wallet-db` information and thus the use case of a custom `neon-wallet-db` node would win over a custom NEO node.

However, I would be changing the title / commit name to `Enable usage with private neon-wallet-db node` because this does allow ppl to use their own hosted node but with the same public chain.

ambethia:
Thanks so much!

PR #64, opened 2017-11-08
[Typescript] Add first typings

ivangabriele:


snowypowers:
Great work! Will review once i get some time. This will take some time xD being on `dev` too much can't remember how `master` looks like now

ivangabriele:
Thank you @snowypowers ! Indeed I noticed the dev branch but didn't take the time to check. It's good to know, especially since I was thinking about cleaning/completing some comments and normalize internal references.

If you need any any help on `dev`, don't hesitate to ask me :)

src/transactions/index.d.ts
```diff
@@ -0,0 +1,160 @@
...
+    GAS: BalanceToken
+    NEO: BalanceToken
+    address: string
+    net: Net
+  }
+  type BalanceToken = {
+    balance: number
+    unspent: BalanceTokenCoin[]
+  }
+  type BalanceTokenCoin = {
```
snowypowers: This should be following the JSDoc name of `Coin`. Look at `api.js`. In fact, this is defined in that document instead of this. Might need shuffling. I am also not sure where this lives.

src/transactions/index.d.ts
```diff
@@ -0,0 +1,160 @@
+declare module 'neon-js' {
+  type Balance = {
+    GAS: BalanceToken
+    NEO: BalanceToken
+    address: string
+    net: Net
+  }
+  type BalanceToken = {
```
snowypowers: This can be called `TokenBalance`

src/transactions/index.d.ts
```diff
@@ -0,0 +1,160 @@
...
+  type TransactionInput = {
+    prevHash: string
+    prevIndex: number
+  }
+  type TransactionOutput = {
+    assetId: string
+    scriptHash: string
+    value: number
+  }
+  type TransactionWitness = {
```
snowypowers: Should be just `Witness`

src/wallet.d.ts
```diff
@@ -0,0 +1,58 @@
+declare module 'neon-js' {
+  type Account = {
+    address: string
+    privateKey: string
+    programHash: string
+    publicKeyEncoded: string
+    publicKeyHash: string
+  }
+
+  function addressToScriptHash(
```
snowypowers: i think this is wrong. there is only 1 arg when looking at source code xD

src/transactions/index.d.ts
ivangabriele: Done

src/transactions/index.d.ts
ivangabriele: Done. Indeed I was afraid that some collusion could happen later.

src/transactions/index.d.ts
ivangabriele: Done

src/wallet.d.ts
ivangabriele: Good catch :)

ivangabriele:
I also all the errors that I didn't see during my first try (because I first wrote the types straight in the dependency to live-test them).

PR #66, opened 2017-11-09
Add Unicode Normalization Form C to NEP2 implementation

ixje:
To match NEP-2 more closely


snowypowers:
I will cherrypick this and integrate this into `dev`

PR #69, opened 2017-11-13
Fix script lengths to use correct formatting

RavenXce:
Script lengths should use standard uint8 / uint16 formatting. 
The current serialization breaks when the invocation script is of a small length (e.g. zero length gives `0` instead of `00`).


notatestuser:
They're actually variable length ints. Should have been fixed in #56. Haven't tried a zero length script though.

RavenXce:
Oh I didn't see that commit, thanks!

PR #78, opened 2017-11-19
Little endian var ints + tests

### LINKED ISSUE #77: Variable length int endianness
I thought `num2VarInt` was doing the right thing until my invocations started failing at the [most inconvenient time](https://git.io/vFHEo). After debugging the node I realised that it was trying to read too much of the byte array, failing on GAS cost deserialization because there was no buffer left. Swapping the bytes around fixed it and the txes went through fine.

[This](https://git.io/vFHEX) is where the wrong var int was coming through - `fd00fd` should be 253 but was becoming 64768 in the node.

I've noticed that we aren't using var ints for big values right now (mostly short array lengths) so I wonder if this was the wrong endianness all along and we are just missing a `reverseHex`.

notatestuser:
fixes #77

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -67,13 +67,16 @@ export const int2hex = mNumber => {
 }
 
 /**
- * Converts a number to a hexstring of a suitable size
+ * Converts a number to a big endian hexstring of a suitable size, optionally little endian
  * @param {number} num
- * @param {number} size - The required size in chars, eg 2 for Uint8, 4 for Uint16. Defaults to 2.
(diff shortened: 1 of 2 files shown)


src/utils.js
```diff
@@ -101,14 +104,20 @@ export const fixed82num = (fixed8) => {
  * @returns {string} hexstring of the variable Int.
  */
 export const num2VarInt = (num) => {
+  if (typeof num !== 'number') throw new Error('VarInt must be numeric')
+  if (num < 0) throw new RangeError('VarInts are unsigned (> 0)')
```
notatestuser: Have noticed the error message should read ">= 0". Not a big deal.

PR #80, opened 2017-11-22
Utils unit tests

notatestuser:
Notes:
* Renamed fixed82num to fixed8hex2num, kept the original export too but added a depreciation warning
* Same as above for num2fixed8/num2fixed8hex
* Small consistency edits

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -66,7 +66,7 @@ export const ab2hexstring = arr => {
 export const str2hexstring = str => ab2hexstring(str2ab(str))
 
 /**
- * convert an integer to hex and add leading zeros
+ * convert an integer to big endian hex and add leading zeros
  * @param {number} num
  * @returns {string}
  */
--- a/src/utils.js
+++ b/src/utils.js
@@ -4,18 +4,22 @@ import { SHA256, RIPEMD160, enc } from 'crypto-js'
  * @param {arrayBuffer} buf
  * @returns {string} ASCII string
  */
-export const ab2str = buf => { return String.fromCharCode.apply(null, new Uint8Array(buf)) }
+export const ab2str = buf =>
+  String.fromCharCode.apply(null, new Uint8Array(buf))
 
 /**
  * @param {string} str - ASCII string
  * @returns {arrayBuffer}
  */
 export const str2ab = str => {
-  let bufView = new Uint8Array(str.length)
-  for (let i = 0, strLen = str.length; i < strLen; i++) {
-    bufView[i] = str.charCodeAt(i)
+  if (typeof str !== 'string') {
+    throw new Error('str2ab expects a string')
   }
-  return bufView
+  const result = new Uint8Array(str.length)
+  for (let i = 0, strLen = str.length; i < strLen; i++) {
+    result[i] = str.charCodeAt(i)
+  }
+  return result
 }
 
 /**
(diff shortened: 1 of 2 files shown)


notatestuser:
Other notes:
* Renamed fixed82num to fixed8hex2num, kept the original export too but added a depreciation console.warn
* Same as above for num2fixed8hex

src/utils.js
```diff
@@ -74,40 +86,56 @@ export const int2hex = mNumber => {
...
- * Converts a number to a Fixed8 format string
+ * Converts a number to a Fixed8 format hex string
  * @param {number} num
+ * @param {number} size output size in hex chars
  * @return {string} number in Fixed8 representation.
  */
-export const num2fixed8 = (num) => {
-  const hexValue = Math.round(num * 100000000).toString(16)
-  return reverseHex(('0000000000000000' + hexValue).substring(hexValue.length))
+export const num2fixed8hex = (num, size = 16) => {
```
snowypowers: Fixed8 is a string format that represents a number with built in accuracy of 8 decimals places. The length of the string is dependent on the number of of trailing zeroes but it can always be represented as a length of 16 characters (8 bytes).

Is there any case where you will enforce a length that is not 16? I do think that there is no need for `fixed8hex` though because there isn't a different representation used in the system except by hexstring

tests/utils.js
```diff
@@ -1,7 +1,214 @@
...
+
+  describe('int2hex', () => {
+    it('throws if non-number', () => {
+      (() => int2hex('01')).should.throw()
+    })
+    it('converts a number to its hex form', () => {
+      const actual = int2hex(0x00)
+      actual.should.eql('00')
+    })
+    it('converts a number to its hex form (max)', () => {
```
snowypowers: This isnt correct. `int2hex` can convert any size. The biggest difference between this and `num2hexstring` is that for `int2hex` you do not need to constraint the size. Thus, `int2hex` is useful for converting arbitrary data but `num2hexstring` requires an upper limit.
```
int2hex(88888) = '015b38' // Doesnt care abt size, just ensures that there is a leading zero if there length is odd.
num2hexstring(88888, 4) = '5b38' // Cares and truncates to length
```

src/utils.js
notatestuser: `fixed8` to me refers to the base 10 form and not the hex form, this is just what I gathered through digging through the neo sources and my initial impressions upon reading this code. we can roll that back if you disagree - it's not a big deal.

I can't think of a situation where we would specify a length other than 16 in neon-js but some contracts may be able to use it. For example, my SC puts a GAS amount in 5 bytes for a maximum of ~5497.5 GAS - (2^40)/2. This is for optimizing storage space if it is already known that it won't exceed that number.

notatestuser: On this note, it would be good to open up discussion about whether we should always specify `size` in bytes and avoid confusion of hex/byte lengths

snowypowers: I do know the NEO nodes return shaved fixed8 (fixed8s with leading zeroes shaved off). 

`Fixed8` to me was always an unreadable encoded form so hex for me. If i wanted base10, I will be using `number`? On a side note, Im not sure if `number` can handle the maximum value of `Fixed8`.

Flexibility is welcomed  for the method so i am ok. It also doesnt break existing implementations.

`size` wise, I think I will lean towards byte lengths so that its clear coming from C#. Hexstrings is just chosen to be the main representation of data used in this repo.

notatestuser: Alright, I could make the changes to use byte sizes in the utils methods.

notatestuser: Will do that as a separate PR. ^

snowypowers:
Merged in but I will change `num2fixed8hex` back to `num2fixed8`. This repo is unlikely to see `fixed8` in a different format other than a hexstring. 

The byte update can be done seperately.

PR #98, opened 2017-12-09
catch getTokenBalance error when using an address with no balance

ixje:
the `stack[0].value` returns ‘’ when using an address with no balance
causing it to throw “fixed8hex must be hex” in `fixed82num()`.


ixje:
@snowypowers note that the test failure is unrelated to my commit. It actually already failed when just pulling the latest release.

snowypowers:
no worries I will look into the tests. Good catch.

PR #101, opened 2017-12-11
Using byte lengths instead of hexstring lengths

notatestuser:
As discussed in #80

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -87,13 +87,15 @@ export const int2hex = num => {
 /**
  * Converts a number to a big endian hexstring of a suitable size, optionally little endian
  * @param {number} num
- * @param {number} size - The required size in hex chars, eg 2 for Uint8, 4 for Uint16. Defaults to 2.
+ * @param {number} size - The required size in bytes, eg 1 for Uint8, 2 for Uint16. Defaults to 1.
  * @param {boolean} littleEndian - Encode the hex in little endian form
  * @return {string}
  */
-export const num2hexstring = (num, size = 2, littleEndian = false) => {
+export const num2hexstring = (num, size = 1, littleEndian = false) => {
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -87,13 +87,15 @@ export const int2hex = num => {
...
+ * @param {number} size - The required size in bytes, eg 1 for Uint8, 2 for Uint16. Defaults to 1.
  * @param {boolean} littleEndian - Encode the hex in little endian form
  * @return {string}
  */
-export const num2hexstring = (num, size = 2, littleEndian = false) => {
+export const num2hexstring = (num, size = 1, littleEndian = false) => {
   if (typeof num !== 'number') throw new Error('num must be numeric')
   if (num < 0) throw new RangeError('num is unsigned (>= 0)')
+  size = size * 2
+  if (size % 2 !== 0) throw new RangeError('size must be a multiple of 2')
```
snowypowers: i think the error msg should be about it being an integer. If it is an integer, it will always be multiple of 2 as you multipled it by 2.

notatestuser: Good point, it's essentially acting as an integer check there. I'll change that check/message

snowypowers: mm the rest looks good, will merge once i see this change

notatestuser:
Should be careful with this change if we think dependent apps might be consuming the utils directly.

snowypowers:
good point. will ask around. v3? ugh dont feel like i should be jumping the majors so fast hahaaha

notatestuser:
If we're following semantic versioning then yeah it should be v3 since those functions were exported. 

However now that everyone uses a lockfile with npm/yarn nowadays and because this is technically still just the library for neon-wallet a minor jump could be okay imo.

PR #110, opened 2017-12-27
Added docs build commands to readme

notatestuser:
Because it's not so obvious

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,15 @@ npm run build:dev
 npm run test
 ```
 
+### Docs
+
+```bash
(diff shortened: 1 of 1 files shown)


snowypowers:
just a qn... did you read the `source/readme.md` file?

notatestuser:
Actually no missed it. But why is that folder called `source`? Also it does not seem to install `sphinx_rtd_theme` which is required.

snowypowers:
The folder name was an artifact that i took over. 

Oh i missed out on the theme.

PR #113, opened 2017-12-27
Added getPrices function for retrieving multiple CoinMarketCap symbols

mhuggins:
I was looking into the possibility of fetching prices for multiple currencies at the same time from within neon-wallet.  It looks like CoinMarketCap exposes an endpoint for fetching all tickers, which we can then filter the results afterwards.  This exposes a new `cmc.getPrices` function to achieve that.

This PR includes changes to files in the `lib` folder that I assume were updated when I ran `yarn run test`, but I'm not sure if they should be included with this PR.


mhuggins:
The test failures on this seem to be unrelated, but maybe something that changed in the `lib` folder broke them...?

> 1) RPC methods test block height:
>    Error: socket hang up
> 
> 2) RPC methods test storage:
>    Error: socket hang up

snowypowers:
no worries its not your fault. seed1 (which most of the default tests points to) is down thats why. The code looks promising, will spend more time  checking it out later tonight.

snowypowers:
sorry @mhuggins re-open this pointing to `dev` instead of `master`

PR #118, opened 2017-12-29
Added doSendAssets for sending to multiple recipients

mhuggins:
This adds a new `doSendAssets` function that behaves similar to the existing `doSendAsset` function, but it allows sending to multiple recipients instead of just one.  I left the original `doSendAsset` to prevent breaking existing implementations.

Some thoughts/questions:

* The test I added is being skipped simply because the other tests around sending assets are currently being skipped as well.  I don't have confidence in what I built because of that, and I'm wondering why the existing tests are being skipped.
* I'm not familiar with the structure for generating documentation based upon the inline comments.  I included `@param {{address: {NEO: number, GAS: number}}} addressAssetAmounts`, but I'm not sure if "address" will be treated literally when that's intended to represent a dynamic key.
* Should the `lib` folder be generated and included in the repo as it is today?  I see that one of the changes in there is replacing "C:\\Users\\JunXiang\\Desktop\\neon-js" with "/Users/mhuggins/Development/neo/neon-js", which seems odd.


snowypowers:
I am trying to transit the way we send assets from `neondDB/doSendAssets` to `core/sendAsset`. The new methods uses an intent based system which supports sending to multiple addresses.

Do have a look and see if it suffices.

mhuggins:
Thanks @snowypowers, I'll take a close look at those. 😄

PR #120, opened 2017-12-30
[WIP] Updated sendAssets to support NEP5 tokens

mhuggins:
This updates the existing `sendAssets` core API function to accepts a new `config.transfers` array for sending tokens in addition to assets.  It also exposes a new `makeTransfer` helper function (similar to `makeIntent`) for creating the transfer script outputs needed to generate the invocation script.

Example:

```js
const fromAddress = 'AP4AswDCgRCRF7Z6oMQMrFBzTuRgbeiQNF'
const toAddress = 'ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW'
const RPX = 'ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9'
const DBC = 'b951ecbbc5fe37a9c280a76cb0ce0014827294cf'

sendAsset({
  net: 'MainNet',
  address: fromAddress
  privateKey: 'foo',
  transfers: [
    makeTransfer({ [RPX]: 1500 }, toAddress, fromAddress),
    makeTransfer({ [RPX]: 1500, [DBC]: 9750 }, toAddress, fromAddress),
  ],
  intents: [
    makeIntent({ NEO: 5, GAS: 0.1 }, toAddress)
  ]
})
```

### DIFF
--- a/src/api/core.js
+++ b/src/api/core.js
@@ -166,9 +166,9 @@ export const makeIntent = (assetAmts, address) => {
  * @param {object} config - Configuration object.
  * @param {string} config.net - 'MainNet', 'TestNet' or a neon-wallet-db URL.
  * @param {string} config.address - Wallet address
- * @param {string} [privateKey] - private key to sign with. Either this or signingFunction is required.
- * @param {function} [signingFunction] - An external signing function to sign with. Either this or privateKey is required.
- * @param {TransactionOutput[]} intents - Intents.
+ * @param {string} [config.privateKey] - private key to sign with. Either this or signingFunction is required.
+ * @param {function} [config.signingFunction] - An external signing function to sign with. Either this or privateKey is required.
+ * @param {TransactionOutput[]} config.intents - Intents.
  * @return {object} Configuration object.
  */
 export const sendAsset = (config) => {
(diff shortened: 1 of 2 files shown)


src/api/core.js
```diff
@@ -161,23 +162,77 @@ export const makeIntent = (assetAmts, address) => {
...
+
+/**
+ * Helper method to convert a TokenAmounts object to transfer invocation scripts (ScriptOutput[]).
+ * @param {object} tokenAmts - Token Amounts
+ * @param {number} tokenAmts.SCRIPT_HASH - Amt of [SCRIPT_HASH] token to send. (TODO: better way to write this for docs?)
+ * @param {string} toAddress - The address to send to.
+ * @param {string} fromAddress - The address to send from.
+ * @return {ScriptOutput[]} ScriptOutput
+ */
+export const makeTransfer = (tokenAmts, toAddress, fromAddress) => {
```
mhuggins: I don't like the idea of passing `fromAddress` here since it's already part of the config, it seems redundant.  I think I can drop it from here and have this return some other new set of objects instead of `ScriptOutput[]`.  Then `createTransferTx` or `createTransferScript` can be responsible for constructing `args`, which includes the representation of the from address.

src/api/core.js
```diff
@@ -161,23 +162,77 @@ export const makeIntent = (assetAmts, address) => {
...
  * @return {object} Configuration object.
  */
 export const sendAsset = (config) => {
   return getBalanceFrom(config, neonDB)
-    .then(
-    (c) => c,
-    () => getBalanceFrom(config, neoscan)
-    )
-    .then((c) => createTx(c, 'contract'))
+    .catch(() => getBalanceFrom(config, neoscan))
```
mhuggins: I changed this from `then(noopFn, catchFn)` to just `catch(catchFn)`, any concerns?

snowypowers: This is ok.

snowypowers:
My biggest concern here is that the infrastructure already exists to make building transactions easier. Building another layer on top of the current `core` layer might just be code bloat without adding significant advantages. All these might belong in `neon-wallet` instead where you transform a button click into an `ScriptOutput`, and the mighty send button will work its magic.

I do believe that as a library, this should not be the source of one-click methods but rather, we provide sufficient tools to do that. `doInvoke`, `claimGas` and `sendAsset` are merely examples of how you can utilise the `core` api to build them in your own implementation.

mhuggins:
@snowypowers Thanks for the feedback.  I think that's reasonable.  Everything I need to do this from neon-wallet is being exported, and I can just call `doInvoke` instead of `sendAsset` if I need to include a script for token transfer.

Issue #99, opened 2017-12-10
[Refactor] API Core code to use seperate promise to determine order of APIs


snowypowers:
In `api.core` the methods `sendAsset`, `claimGas` and `doInvoke` chains promises to determine which API to call first. To switch the order, we have to adjust 3 places in code. We would like to refactor such that there is a way to determine the order in a central location for these methods to refer to.

## Specifics

- Create a function that references an array for the API order
- This function returns a promise which calls the given function with the APIs in order, returning the data required by the 3 functions mentioned above.

PR #121, opened 2017-12-30
feat(api): Api switch to toggle between neonDB and neoscan

snowypowers:
An API switch variable that allows toggling between neonDB and neoscan.

This allows us to load balance and ensure we don't keep hitting an endpoint when its down.

Methods:

`api.core.setApiSwitch`: set the apiSwitch to a predetermined number. 0 for 100% neoscan, 1 for 100% neonDB.

`api.core.setSwitchFreeze`: set the switch to freeze or not. Frozen switch will not dynamically react to failing endpoints.


snowypowers:
Resolves #99

PR #122, opened 2018-01-02
fix(tx): core.signTransaction to check for privateKey

snowypowers:
- Implement a simple check for privateKey before signing. Encountered case where WIF was inserted instead and resulted in wrong signature.

- Expanded `Transaction.sign` to accept `Account` and `WIF`

- write tests for `Transaction.sign`


mhuggins:
Looks good to me, thanks!

PR #123, opened 2018-01-02
Fixed params comments to nest config arguments

mhuggins:
This updates the params comments to properly nest some keys under the `config` object that they belong to.  Without this, it makes it look like these are separate function arguments when they are not.


snowypowers:
rebase on latest `dev`

mhuggins:
@snowypowers Rebased.

PR #142, opened 2018-01-15
fix(api): getPrices patch

### LINKED ISSUE #141: getPrices for cmc is not showing all prices
If you try to get the price of RPX, it will not work, because cmc api returns by default only the top 100 results.
Try to look for RPX here: https://api.coinmarketcap.com/v1/ticker/ its not there.

However, if you pass a `?limit=0` to the query, it will appear - https://api.coinmarketcap.com/v1/ticker/?limit=0

Should I make a PR, introducing the limit param? or this should be just the default? (show all)

evgenyboxer:
Fixes #141 

Adds a `limit=0` to the query.
I have also added another test in getPrices which tests the currency variable.

Note - the `limit=0` will also work with getPrice (it won't really affect it)


snowypowers:
I tweaked the interface by renaming it to `Prices` instead of `Price`. Else the rest is good :)

PR #151, opened 2018-01-22
Async encryption

dvdschwrtz:


src/consts.js
```diff
@@ -36,6 +36,13 @@ export const DEFAULT_SCRYPT = {
   size: 64
 }
 
+export const DEFAULT_SCRYPT_ASYNC = {
```
snowypowers: We can just used the old DEFAULT_SCRYPT instead of creating a new one

test/unit/wallet/nep2.js
```diff
@@ -2,13 +2,21 @@ import * as NEP2 from '../../../src/wallet/nep2'
...
+require('babel-polyfill')
+
+describe.only('NEP2', function () {
   const simpleScrypt = {
     cost: 256,
     blockSize: 1,
     parallel: 1,
     size: 64
   }
+  const simpleAsyncScrypt = {
```
snowypowers: we should follow the old scrypt conventions. Or change everything to n,r,p . `dkLen` is not variable (fixed at 64)

test/unit/wallet/nep2.js
```diff
@@ -2,13 +2,21 @@ import * as NEP2 from '../../../src/wallet/nep2'
 import { isNEP2, isWIF } from '../../../src/wallet/verify'
 import testKeys from '../testKeys.json'
 
-describe('NEP2', function () {
+require('babel-polyfill')
+
+describe.only('NEP2', function () {
```
snowypowers: only

src/wallet/nep2.js
```diff
@@ -71,4 +105,38 @@ export const decrypt = (encryptedKey, keyphrase, scryptParams = DEFAULT_SCRYPT)
...
+/**
+ * Decrypts an encrypted key using a given keyphrase under NEP-2 Standard.
+ * @param {string} encryptedKey - The encrypted key (58 chars long).
+ * @param {string} keyphrase - The password will be encoded as UTF-8 and normalized using Unicode Normalization Form C (NFC).
+ * @param {scryptParams} [scryptParams] - Parameters for Scrypt. Defaults to NEP2 specified parameters.
+ * @returns {string} The decrypted WIF key.
+ */
+export const decryptAsync = (encryptedKey, keyphrase, scryptParams = DEFAULT_SCRYPT_ASYNC) => {
+  return new Promise((resolve, reject) => {
+    scryptParams = ensureScryptParams(scryptParams)
```
snowypowers: `ensureScryptParams` ensures the fields cost, parallel and blockSize are filled. This does nothing for you because you are using n,r,p.

test/unit/wallet/nep2.js
dvdschwrtz: ok, will make these changes later. and I will follow the old conventions

Issue #147, opened 2018-01-20
Adding signing for SC to transaction


snowypowers:
In order to manipulate assets from a smart contract address, we need certain addons to a normal `ContractTransaction`. We would like to have this as a single method under the `Transactions` class that appends the necessary addons.

Given an `adminKey` (The privatekey of the admin address) and `contractScriptHash` (the scripthash of the smart contract),

- Add an attribute of type `Script`, data being the scripthash of the `adminKey`.
- Add an signature where:
  - The invocationScript is `00` + standard invocationScript signed by `adminKey`.
  - The verificationScript is `result.script` retrieved from the `getcontractstate` call using the `contractScriptHash`.

## Specifics
- Implement method `Transaction.signAsAdmin` (name is up for discussion)
- This method should have the params: `Transaction.signAsAdmin(privateKey: string, contractScriptHash: string)`
- This method will append the relevant items mentioned above and return the `Transaction` object.

slipo:
We also need something along the lines of ...
```
            attributes: [
              {
                usage: 32,
                data: this.reverseHex(wallet.getScriptHashFromAddress(sourceAddr))
              }
            ],
```

slipo:
The "00" is that way because the script we were testing has 2 arguments. It needs to match the contract's parameter count.

slipo:
A couple more notes:
* The order of the scripts is important. They need to be sorted by script hash, ascending. Otherwise the whole thing fails.
* This isn't only for admins of the script. I think it should be a method called fromContractScriptHash() or something like that which modifies the transaction and adds the attribute + script.

snowypowers:
- Can you elaborate on the order? What do you mean by ascending?
- Is there any order to attributes ( I would suppose none but you can try adding remarks before or after)
- The main purpose is to attach a signature that verifies that the signer is authorised to carry out the action stated in the SC. Whether the method should be in the Transaction or Account Class is a tossup.

slipo:
Sure, essentially I'm doing something like this in my custom code:
```
      if (parseInt(contractScriptHash, 16) > parseInt(sender.scriptHash, 16)) {
        signedTx.scripts.push({
          invocationScript: '00' + signedTx.scripts[0].invocationScript,
          verificationScript: script,
        })
      } else {
        signedTx.scripts.unshift({
          invocationScript: '00' + signedTx.scripts[0].invocationScript,
          verificationScript: script,
        })
      }
```

I believe that's what this is doing:
[...]

PR #158, opened 2018-02-01
feat(api): Support sending assets from contract

slipo:
Solves https://github.com/CityOfZion/neon-js/issues/147

This PR allows you to send assets from a contract. It supports Contract and Invocation transactions (which are used in neo-boa examples such as https://github.com/CityOfZion/neo-boa/blob/master/boa/tests/src/WithdrawTest.py#L100).

I saw very similar functionality for mintTokens. I'm unsure how mintTokens was being used in this case. If we need it to work as before, is there a way to test it? It will be easy to add back in the exact same way as before if needed.

For my implementation I set `invocationScript` to '00' * number of contract parameters which is what the mintTokens version was doing. I did not append the main script's `invocationScript` as the issue says as this does not seem to be needed. The original mintToken implementation was using the contract's script hash in the attribute and I needed to use the sender's. Again, it wasn't clear to me how mintToken was working. What we have here is much more like how neo-python works with `--from-addr` when sending.

Here is some testing code:
https://gist.github.com/slipo/c1542c087bff7e17a067e903f0b24540

Here's a resulting transaction:
https://gist.github.com/slipo/6b2815058652fce6b386267f1b880639

### DIFF
--- a/src/api/core.js
+++ b/src/api/core.js
@@ -1,4 +1,4 @@
-import { Account } from '../wallet'
+import { Account, getScriptHashFromAddress } from '../wallet'
 import { ASSET_ID } from '../consts'
 import { Query } from '../rpc'
 import { Transaction, TransactionOutput, TxAttrUsage } from '../transactions'
@@ -31,8 +31,10 @@ export const sendAsset = config => {
   return loadBalance(getRPCEndpointFrom, config)
     .then(url => Object.assign(config, { url }))
     .then(c => loadBalance(getBalanceFrom, config))
+    .then(c => addAttributesIfExecutingAsContract(c))
     .then(c => createTx(c, 'contract'))
     .then(c => signTx(c))
+    .then(c => attachContractIfExecutingAsContract(c))
     .then(c => sendTx(c))
     .catch(err => {
       const dump = {
@@ -90,10 +92,10 @@ export const doInvoke = config => {
   return loadBalance(getRPCEndpointFrom, config)
     .then(url => Object.assign(config, { url }))
     .then(c => loadBalance(getBalanceFrom, config))
-    .then(c => addAttributesForMintToken(c))
+    .then(c => addAttributesIfExecutingAsContract(c))
     .then(c => createTx(c, 'invocation'))
     .then(c => signTx(c))
-    .then(c => attachInvokedContractForMintToken(c))
+    .then(c => attachContractIfExecutingAsContract(c))
     .then(c => sendTx(c))
     .catch(err => {
       const dump = {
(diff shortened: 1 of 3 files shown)


snowypowers:
yes we need the mintToken functionality. Please update it such that the old functions are retained and make your own new ones. They serve very different functions although they look similar. The `mintToken` function allows the contract to reject the transaction if the token sale is oversubscribed / finished. We are adding features to allow control over SC assets.

Do make a distinction for a smart contract. A normal send asset transaction is called `ContractTransaction` because an address is considered a normal contract.

src/api/core.js
```diff
@@ -219,50 +221,73 @@ export const makeIntent = (assetAmts, address) => {
...
+
+  return false
+}
+
 /**
  * Adds attributes to the override object for mintTokens invocations.
  * @param {object} config - Configuration object.
  * @return {object} Configuration object.
  */
-const addAttributesForMintToken = config => {
```
snowypowers: We want our mintToken functionality.

test/unit/api/core.js
```diff
@@ -157,6 +157,33 @@ describe('Core API', function () {
         })
     })
 
+    it('sign for contract assets', () => {
```
snowypowers: do use SC or smart contract instead of just contract. An normal address is a normal contract.

slipo:
Thank you for the feedback. I've made all changes.

I see that mintTokens is a special case where the assets are coming from the sender's regular address but they want the verification triggered. I've put all original mintTokens code back.

I also have created a sendAssets example:
https://gist.github.com/slipo/b4ff4569ada4a93fce683c3da512e0f2

Please see my question in that code. It seems to me we need to (option 1, shown in example) pass along a scriptHash when doing this form of send assets. Otherwise we need to (option 2) remove this:

```
    if (config.address !== acct.address && !sendingAddressIsSmartContract(config)) {
      return Promise.reject(
        new Error('Private Key and Balance address does not match!')
      )
    }
```

Or a third option is to add a new `config.sendingFromSmartContract = true`.  I would personally prefer options 1 or 3 because I think that this isn't the most common usage and that error message could be useful for people.

snowypowers:
yes, I think adding a `sendingFromSmartContract` field would be the option here. We do not want to deviate from the original purpose of sending assets from a personal account. Another solution could be to create a new method called `sendAssetsFromSC` in order to differentiate. I will accept either methods for now.

slipo:
Done. Looking cleaner to me.

snowypowers:
Thanks for the work! There might be some changes as to where this should live but that should be minor.

PR #168, opened 2018-02-20
Update param.value validation logic to accept 0 and empty string

rockacola:
Take the following for example:

```
const props = {
  scriptHash: 'CONTRACT_HASH',
  operation: 'do_stuff',
  args: [
    { "type": "String", "value": '' },
    { "type": "Integer", "value": 0 }
  ],
}
const script = Neon.sc.createScript(props)
```

Supplied intents meant to be valid but are blocked by `_emitParam` due to its validation check.

I've create a separate private method `_isValidValue()` that will allow us to expand on this logic in the future if needed while keeping `_emitParam()` readable.

### DIFF
--- a/src/sc/ScriptBuilder.js
+++ b/src/sc/ScriptBuilder.js
@@ -82,7 +82,7 @@ class ScriptBuilder extends StringStream {
    */
   _emitParam (param) {
     if (!param.type) throw new Error('No type available!')
-    if (!param.value) throw new Error('No value available!')
+    if (!this._isValidValue(param.value)) throw new Error('Invalid value provided!')
     switch (param.type) {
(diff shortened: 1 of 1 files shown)


snowypowers:
will it be better if we just have this method live outside? it looks like a static method to me. At least `_emitParam` can be called if they really want to but `_isValidValue` doesnt follow the builder pattern.

snowypowers:
i think  put it outside as a method but not part of the class so we dont expose it.

```
class ScriptBuilder {
...
}

const isValidValue = () => {...}

export default ScriptBuilder
```

rockacola:
Thanks for the direction.

Updated the PR with said changes.

PR #171, opened 2018-02-25
wrong documentation: rpc not part of Neon

gigerbytes:


snowypowers:
So this is a murky area depending on how you import...

if you use ES6 imports and Neon is your default import, it will not have the modules.
However, if you use `require`, Neon encompasses everything and `Neon.rpc` is valid.

I should be clearer in my examples and not switch between the two imports too often.

gigerbytes:
Thanks for the clarification

PR #185, opened 2018-03-04
Refactor(typings): Modularize typings & Add descriptions

nickfujita:
As this repo grows, it makes sense to break the typings into modular files rather that a single typing file for each package folder. It will help to separate typing concerns and make it easier to find definitions.

Short descriptions were added for each typing where possible. This will provide users with a short bit of information when hovering over usages in IDEs such as Atom or VSCode.


src/api/typings/coinmarketcap.d.ts
```diff
@@ -0,0 +1,9 @@
+interface Prices {
+  [key: string]: number
+}
+
+/** Returns the price of coin in the symbol given */
+export function getPrice(coin?: string, currency?: string): Promise<number>
```
snowypowers: Following your convention, this should get a `DEPRECATED` tag

src/api/typings/core.d.ts
```diff
@@ -0,0 +1,95 @@
...
+  NEO: number
+  block_index: number
+  gas_sent: boolean
+  neo_sent: boolean
+  txid: string
+}
+
+export interface apiConfig {
+  net: net
+  address: string
```
snowypowers: The introduction of `account` would mean this is optional. Yes, its getting messy. I can feel a refactor coming up.

src/api/typings/neonDB.d.ts
```diff
@@ -0,0 +1,75 @@
...
+): Promise<RPCResponse>
+
+/** DEPRECATED: use claimGas instead */
+export function doClaimAllGas(
+  net: net,
+  publicKey: string,
+  signingFunction: signingFunction
+): Promise<RPCResponse>
+
+/** Call mintTokens for RPX */
```
snowypowers: All these `doX` functions are deprecated

src/transactions/typings/txAttrUsage.d.ts
```diff
@@ -0,0 +1,41 @@
+/** Enum for TransactionAttributeUsage */
+export interface TxAttrUsage {
```
snowypowers: Not an expert on typings but I recall Typescript having enums. Is `interface` correct here?

src/wallet/ClaimItem.js
```diff
@@ -0,0 +1,32 @@
+import { Fixed8 } from '../utils'
```
snowypowers: Might be worth to shift these into a components folder so that the `wallet` folder exposes the top level stuff. These functions aren't meant to be used regularly by consumers.

src/wallet/typings/Account.d.ts
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * This allows for simple utilisation and manipulating of keys without need the long access methods.
+ * Key formats are derived from each other lazily and stored for future access.
+ * If the previous key (one level higher) is not found, it will attempt to generate it or throw an Error if insufficient information was provided (eg. trying to generate private key when only address was given.)
+ * NEP2 <=> WIF <=> Private => Public => ScriptHash <=> Address
+ */
+export class Account {
+  constructor(str: string|object)
+
+  WIF: string
```
snowypowers: I believe there is an `encrypted` property now.

src/wallet/typings/Balance.d.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/** Describes the coins found within an Account. Look up various balances through its symbol. For example, NEO or GAS. */
+export class Balance {
+  constructor(bal?: Balance)
+
+  /** The address for this Balance */
+  address: string
+
+  /** The network for this Balance */
+  net: 'MainNet' | 'TestNet'
```
snowypowers: This can be any string too.

src/wallet/typings/Wallet.d.ts
```diff
@@ -0,0 +1,62 @@
+export interface WalletFile {
```
snowypowers: Might be consistent to rename this to `WalletLike`

src/api/typings/core.d.ts
nickfujita: yes, this was actually what I was going to suggest next. Passing around this mutating object is getting a bit unwieldy. Was probably more convenient at the start, but now that it's starting to mature and settle a bit, might make sense to make the functions that use this object have specific inputs and outputs

nickfujita: just updated to optional for the sake of this PR tho

src/api/typings/coinmarketcap.d.ts
nickfujita: added

src/api/typings/neonDB.d.ts
nickfujita: added `DEPRECATED` for `doX` methods

src/transactions/typings/txAttrUsage.d.ts
nickfujita: updated to enum along with sc.opCode

src/wallet/typings/Account.d.ts
nickfujita: added

src/wallet/typings/Balance.d.ts
nickfujita: updated to use same type from `api.core`

src/wallet/typings/Wallet.d.ts
nickfujita: updated to WalletLike

src/wallet/ClaimItem.js
nickfujita: these `components` are currently exported from the wallet package folder
https://github.com/CityOfZion/neon-js/blob/master/src/wallet/index.js#L44

I can place them into a `components` folder, but removing them from the export would be a breaking change. If this is desired, I can updated for this.

nickfujita:
updated typings to differentiate between `Class` and `ClassLike`, also merged in and added typings for messaging methods

Said007:
@nickfujita Nice I guess I don't need to write any docs for the message :)

nickfujita:
@Said007 👍 well these are just the typings with quick descriptions, the docs are a separate initiative:
https://github.com/CityOfZion/neon-js/tree/master/docs/wallet

Said007:
Ops your right. I thought maybe the typings will be used to generate the docs.

src/wallet/ClaimItem.js
snowypowers: Place them into a folder but flatten the exports at the `wallet` level. I would like your opinion on this though. Does having too many files clutter the `wallet` folder or is it fine?

src/wallet/ClaimItem.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @typedef ClaimItem
+ * @property {Fixed8} claim - Amt of gas claimable
+ * @property {string} txid - Transaction hash of the originaating coin
+ * @property {number} index - Index of coin in the output array
+ * @property {number} value - Amount of NEO involved.
+ * @property {Fixed8} [start] - Starting block. Optional.
+ * @property {Fixed8} [end] - Ending block. Optional.
+ */
+export default class ClaimItem {
```
snowypowers: This is considered a breaking change if we do not provide a non-new method to go with it.

nickfujita: Reverted ClaimItem back to original format.

src/wallet/ClaimItem.js
nickfujita: I think it makes since to keep the components together, so put them all into a folder. Good compromise between getting too many files in the wallet folder and having all components in one file.

src/wallet/typings/message.d.ts
```diff
@@ -0,0 +1,9 @@
+
+/** Converts signatureHex to a signature object with r & s. */
+export function getSignatureFromHex(signatureHex: string): {}
```
snowypowers: this isnt exported

src/wallet/typings/Wallet.d.ts
```diff
@@ -0,0 +1,54 @@
...
+  decryptAll(keyphrase: string): boolean[]
+
+  /** Attempts to encrypt Account at index in array. */
+  encrypt(index: number, keyphrase: string): boolean
+
+  /** Attempts to encrypt all accounts with keyphrase. */
+  encryptAll(keyphrase: string): boolean[]
+
+  /** Export this class as a string */
+  export(): string
```
snowypowers: I believe this returns `WalletLike`

snowypowers:
Im sorry that im requesting for so many changes.... but let's not do such big PRs again, its taking a bit too long to make sure everything is right.

src/wallet/typings/message.d.ts
nickfujita: removed

src/wallet/typings/Wallet.d.ts
nickfujita: updated comments as well.

nickfujita:
Don't be sorry about requesting changes. That's what PRs are for. But yes, next time I will section these PRs into more manageable chunks.

snowypowers:
phew i think thats most of it. Thanks for the work!

PR #198, opened 2018-03-17
fix typings error and added typings script check

wmira:


package.json
```diff
@@ -61,7 +62,8 @@
     "build:docs": "cd docs && sphinx-versioning -g ../ -l ./conf.py build docs ../docs/_build/html",
     "deploy:docs": "cd docs && sphinx-versioning -g ../ -l ./conf.py push docs gh-pages .",
     "rebuild": "rm -rf ./node_modules && npm install && ./node_modules/.bin/webpack",
-    "prepublishOnly": "npm run lint && npm run build:prod"
+    "prepublishOnly": "npm run lint && npm run build:prod",
+    "compile-typings": "tsc src/**/*.d.ts"
```
snowypowers: I tried this locally on windows and it din't work. I have `typescript 2.7.2` in my global npm. Instead, `tsc src/index.d.ts` works for me.

wmira: hmm weird i assume should work same on windows. here is on my ubuntu.
wmira@wmira-All-Series:/mnt/data/dev/pullreq/neon-js$ npm run compile-typings
> @cityofzion/neon-js@3.4.4 compile-typings /mnt/data/dev/pullreq/neon-js
> tsc src/**/*.d.ts

wmira@wmira-All-Series:/mnt/data/dev/pullreq/neon-js$ 

will get on the windows box and check why

snowypowers: any updates?

wmira: just tried it on windows now. Yeah weird that it is throwing error. Trying to see how to resolve

wmira:
@snowypowers i have added tsconfig.json to point to the folder and remove it from the tsc command from the npm script.

package.json
```diff
@@ -61,7 +62,8 @@
     "build:docs": "cd docs && sphinx-versioning -g ../ -l ./conf.py build docs ../docs/_build/html",
     "deploy:docs": "cd docs && sphinx-versioning -g ../ -l ./conf.py push docs gh-pages .",
     "rebuild": "rm -rf ./node_modules && npm install && ./node_modules/.bin/webpack",
-    "prepublishOnly": "npm run lint && npm run build:prod"
+    "prepublishOnly": "npm run lint && npm run build:prod",
+    "compile-typings": "tsc"
```
snowypowers: I think a more appropriate name here would `test:typings` as the use case is probably largely during dev process. It doesn't actually compile and produce anything. We can add this to the `prepublishOnly` script.

wmira:
@snowypowers renamed test:tyings and added call on prepublishOnly. prepublishOnly fails on typings error

snowypowers:
alright looks good to me, I think there is some stuff I gotta do to improve CI but I will merge this up and deal with it myself.

PR #202, opened 2018-03-20
handle decimal sizes other than 8 when converting to ByteArray

hal0x2328:
neon-js currently does not properly handle sending tokens with decimal sizes other than 8. For example, when sending 1 RHT (a zero-decimal token), neon-js assumes a fixed8 size and converts the number "1" to "100000000" when creating the byte array to add to the contract invocation script. The current workaround for wallets using neon-js is for the user to send 0.00000001 RHT for every 1 RHT intended, a less-than-ideal situation

This fix adds a decimal argument to the byteArray function (this argument is already passed to the function by the latest neon-wallet versions) and pre-divides the value according to the number of decimals before it is passed to num2fixed8 at the end of the function, the end result being a byte array with the correct integer value in fixed8 format.

This fix also checks the precision of the value being sent to enforce that no more than the allowed number of decimals are specified, throwing an error (for example, if someone tries to send fractional values of RHT) instead of simply rounding the value to the allowed precision (which could lead to confusion and unexpected loss of funds if users do not know the token's decimals).

### DIFF
--- a/src/sc/ContractParam.js
+++ b/src/sc/ContractParam.js
@@ -64,11 +64,19 @@ class ContractParam {
    * @param {string} format - The format that this value represents. Different formats are parsed differently.
    * @return {ContractParam}
    */
-  static byteArray (value, format) {
+  static byteArray (value, format, decimals = 8) {
     if (format) format = format.toLowerCase()
     if (format === 'address') {
       return new ContractParam('ByteArray', reverseHex(getScriptHashFromAddress(value)))
     } else if (format === 'fixed8') {
+      if (!isFinite(value)) throw new Error(`Input should be number!`)
+      var e = 1, p = 0
+      while (Math.round(value * e) / e !== value) { 
+        e *= 10
+        p++ 
+      }
+      if (p > decimals) throw new Error(`Input value exceeds maximum precision!`)
+      value = value / Math.pow(10, 8 - decimals)
       return new ContractParam('ByteArray', num2fixed8(value))
     } else {
       return new ContractParam('ByteArray', value)
--- a/src/sc/ContractParam.js
+++ b/src/sc/ContractParam.js
@@ -1,4 +1,4 @@
-import { num2fixed8, reverseHex } from '../utils'
+import { num2fixed8, reverseHex, Fixed8 } from '../utils'
 import { getScriptHashFromAddress, isAddress } from '../wallet'
 
 /**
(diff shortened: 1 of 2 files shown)


src/sc/ContractParam.js
```diff
@@ -64,11 +64,19 @@ class ContractParam {
    * @param {string} format - The format that this value represents. Different formats are parsed differently.
    * @return {ContractParam}
    */
-  static byteArray (value, format) {
+  static byteArray (value, format, decimals = 8) {
```
snowypowers: These arguements will not match the method `byteArray`. I would suggest `(value, format, ...args)` so we can be flexible in receiving arguments.

src/sc/ContractParam.js
```diff
@@ -64,11 +64,19 @@ class ContractParam {
...
       return new ContractParam('ByteArray', reverseHex(getScriptHashFromAddress(value)))
     } else if (format === 'fixed8') {
+      if (!isFinite(value)) throw new Error(`Input should be number!`)
+      var e = 1, p = 0
+      while (Math.round(value * e) / e !== value) { 
+        e *= 10
+        p++ 
+      }
+      if (p > decimals) throw new Error(`Input value exceeds maximum precision!`)
+      value = value / Math.pow(10, 8 - decimals)
```
snowypowers: These arithmetic calculations may introduce floating point errors. Please use `Fixed8` class (which is just inherited from BigNumber class) to ensure there is no calculation errors.

src/sc/ContractParam.js
```diff
@@ -62,13 +62,29 @@ class ContractParam {
...
+      var e = 1
+      var p = 0
+      while (Math.round(value * e) / e !== value) {
+        e *= 10
+        p++
+      }
+      if (p > decimals) throw new Error(`Input value exceeds maximum precision!`)
+      value = new Fixed8(value)
+      let divisor = new Fixed8(Math.pow(10, 8 - decimals))
+      value /= divisor
```
snowypowers: You cannot use normal arithmetic functions with Fixed8. You will have to use `value = value.div(divisor)`.

src/sc/ContractParam.js
```diff
@@ -62,13 +62,29 @@ class ContractParam {
...
+      if (args.length === 1) {
+        decimals = args[0]
+      }
+      if (!isFinite(value)) throw new Error(`Input should be number!`)
+      var e = 1
+      var p = 0
+      while (Math.round(value * e) / e !== value) {
+        e *= 10
+        p++
+      }
```
snowypowers: A `while` loop looks very dangerous here. I would suggest just using the Fixed8 class to determine if your precision is exceeded.
```js
const fixed8Value = new Fixed8(value)
const adjustedValue = fixed8Value.times(Math.pow(10, 8 - decimals))
const modValue = adjustValue.mod(1)
if (!modValue.isZero()) throw new Error(`wrong precision`)
return adjustedValue
```

test/unit/sc/ContractParam.js
```diff
@@ -51,6 +51,16 @@ describe('ContractParam', function () {
...
+
+    it('fixed8 with zero decimals', () => {
+      const cp = ContractParam.byteArray(1, 'fixed8', 0)
+      cp.should.eql({ type: 'ByteArray', value: '0100000000000000' })
+    })
+
+    it('fixed8 with four decimals', () => {
+      const cp = ContractParam.byteArray(222, 'fixed8', 4)
+      cp.should.eql({ type: 'ByteArray', value: 'e0df210000000000' })
+    })
```
snowypowers: I would like the case for 4 decimals to actually have some decimals in the input.

Can I also have a failure case to ensure the exceed precision logic is sound.

Issue #224, opened 2018-04-18
[bug] Faulty url in network config file


DalderupMaurice:
Due to a faulty url in the network config file, you are unable to connect to the CoZ network.
This can be fixed by changing the url to the correct url.

PR #225, opened 2018-04-18
fix(src): fixed faulty url in network config file

DalderupMaurice:
fix for issue #224


snowypowers:
I will merge this into `dev` and wait till Monday for a release as this is not a major bug and there is a available workaround by overrwriting the network configuration manually.

PR #236, opened 2018-04-30
fix(sc):fix scriptBuilder toScriptParams error parsing nonce

### LINKED ISSUE #235: [Bug] ScriptBuilder toScriptParams cannot parse code-based nonce
I mentioned this in discord, but invocation transactions with code-based nonce in the script hexstring cannot currently be parsed by toScriptParams as the nonce will trigger "Encounter unknown byte" message.

After further research, it looks like code-based nonces put the nonce after the RET op code, which means the nonce would be ignored by the VM anyway. I suggest adding a case for 102 where the rest of the hexstring is ignored. Should be a two-liner.

achen27:
Closes #235

### DIFF
--- a/src/sc/ScriptBuilder.js
+++ b/src/sc/ScriptBuilder.js
@@ -248,7 +248,7 @@ const retrieveAppCall = (sb) => {
         throw new Error(`Encounter unknown byte: ${b}`)
     }
   }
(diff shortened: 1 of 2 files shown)


snowypowers:
Thanks for the contribution! Code looks simple enough. Can I have a test case for each scenario?

PR #269, opened 2018-06-27
just reminding void return is ff00

igormcoelho:
Fellows, it seems that return code for Void (ff) is not passed as `ff`, but as `ff00`, because it is its biginteger representation (see: https://github.com/neo-project/neo/pull/253). For regular types <= 16, it continues the same, a single hex, because it is the same as bigint representation.


vncoelho:
Great insight, Igor. You have eagle birds.

PR #281, opened 2018-07-13
Token balances

dvdschwrtz:


src/api/nep5.js
```diff
@@ -81,6 +81,61 @@ export const getTokenBalance = (url, scriptHash, address) => {
...
+            log.error(`single call in getTokenBalances failed with : ${e}`)
+          }
+        }
+        return acc
+      })
+      .catch(err => {
+        log.error(`getTokenBalances failed with : ${err.message}`)
+        throw err
+      })
+  }, {})
```
snowypowers: This is pretty complicated logic that should be split into more methods. I am of the opinion that if one token fails to parse, we should fail the entire method. Else, we will tend not to notice this failure when most of the tokens return.

test/integration/api/nep5.js
```diff
@@ -47,6 +47,24 @@ describe('Integration: API NEP5', function () {
...
+      ],
+      testKeys.b.address
+    ).then(result => {
+      result.RPX.should.equal(0)
+      result.RHTT4.should.equal(2)
+      result.LWTF.should.equal(0.00000786)
+      result.NXT.should.equal(0)
+    })
+  })
+
```
snowypowers: Need a test to test the logic for more than 6 tokens.

test/integration/api/nep5.js
```diff
@@ -47,6 +47,24 @@ describe('Integration: API NEP5', function () {
...
+        CONTRACTS.TEST_RHTT4,
+        CONTRACTS.TEST_LWTF,
+        CONTRACTS.TEST_NXT
+      ],
+      testKeys.b.address
+    ).then(result => {
+      result.RPX.should.equal(0)
+      result.RHTT4.should.equal(2)
+      result.LWTF.should.equal(0.00000746)
+      result.NXT.should.equal(0)
```
snowypowers: Let us check that the results returned should be have the same length as the array of scripthashes. We cannot assert that the balances will always remain the same else the test will be too fragile.

dvdschwrtz: Ok cool. will change it shortly.

PR #289, opened 2018-08-12
Check if message is already a Hex

vncoelho:


snowypowers:
This might a little dangerous. The intent of `signMessage` is not to sign hexstrings but rather human readable messages. It might be totally possible that my message looks like a hexstring but I would want the behavior to be consistent there.

I think what is better is to have a `verify` function that verifies signatures for transaction signatures instead. This is already done in v4.

vncoelho:
Hi, Snowy, thanks for the quick reply.

In fact, the `signMessage` can behave nowadays.
I see it more dangerous to stay as it is, because someone can sign a Hex Message and it will be converted to Hex again.
Thus, the only think I did was to check if the message is already a Hex, there is not dangerous implications.

Following the same reasoning, the `verify` do the same.

This is an urgent thing that we need.

vncoelho:
closed via https://github.com/CityOfZion/neon-js/commit/5c762c5dc910a99c794c7103e3a06c54703cd68a#diff-021c7dc58adbef2486852c2ce5a36d72R148

A throw exception should still be included in signMessage and VerifyMessage functions it they are not string and are already a Hex.

vncoelho:
closed via https://github.com/CityOfZion/neon-js/commit/5c762c5dc910a99c794c7103e3a06c54703cd68a#diff-021c7dc58adbef2486852c2ce5a36d72R148

Issue #295, opened 2018-08-15
RPC request have a hardcoded timeout of 5000ms w/ no override


nickfujita:
The current RPC request client has a hardcoded timeout of 5000ms, and return a failure after such time. I am not sure what the previous use case was for having this timeout, but when executing contract invocations, there have been cases where nodes have taken more than 5000ms to process a request. While this is not a common occurrence, it has happened. In those cases, since the timeout artificially reported back a request failure, and the requests have actually succeeded. This is problematic, as it can cause an application or the downstream user to think that the transaction failed to be sent. When in reality, the request was actually accepted by the RPC node, it just took longer than the timeout.

The real world implication is that if a user tries to perform a NEP-5 token transfer, but the request comes back as a failure due to this timeout, the user will most likely retry to send their transaction. The result is that the user will sent 2 transfers of a desired amount, and and double the amount they originally intended to. It will then be up to the receiving party to refund the amount at their discretion.

snowypowers:
There is a timeout setting within `settings.timeout.rpc` that governs the timeout settings for all rpc calls. You can use that to tweak it for all calls.

During timeout, an timeout error will be thrown so it is clear that an timeout occurred. I guess we can raise the defaults to a much higher default like 30s instead.

nickfujita:
Ah, ok so you can just set this value once on init, and it should carry over. Sry, missed this, as the settings were not on the export typings from the package.

I updated the default timeout and the typings in the PR.

PR #296, opened 2018-08-15
Add RPC request timeout override option

nickfujita:
Issue: https://github.com/CityOfZion/neon-js/issues/295

This change provides neon-js users an option to increase the RPC request timeout at their discretion.

### DIFF
--- a/src/rpc/query.js
+++ b/src/rpc/query.js
@@ -328,7 +328,7 @@ export const queryRPC = (url, req, config = {}) => {
     headers: {
       'Content-Type': 'application/json'
     },
-    timeout: timeout.rpc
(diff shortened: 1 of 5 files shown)


src/rpc/query.js
```diff
@@ -328,7 +328,7 @@ export const queryRPC = (url, req, config = {}) => {
     headers: {
       'Content-Type': 'application/json'
     },
-    timeout: timeout.rpc
+    timeout: config.timeout || timeout.rpc
```
snowypowers: You dont have to do this because config is passed in afterwards.

nickfujita:
Updated based on feedback in the issue.

Issue #207, opened 2018-03-27
[Feature] Multi Sig Support


snowypowers:
- Ability to create a multi-sig address
- Ability to discern a multi-sig address if possible

CHAOWEICHIU:
Just wondering,
Is this multi sig support going to follow the spec of Bitcoin multi sig implementation from https://bitcoin.org/en/developer-guide#multisig?

snowypowers:
I really have no clue, someone gotta dive into the code more.. (that's why i din't specify any more info)

jeroenptrs:
@snowypowers I'm testing out the creation of multisigs, porting it from `neo-python`. You can assign this to me if you like 😄

PR #301, opened 2018-08-25
feat(neon-core): multisig Account and Witness

snowypowers:
- Account.createMultiSig
- Witness.buildMultiSig
- change test rpc url


snowypowers:
Closes #207

PR #303, opened 2018-08-28
Fix createContractTx typings

nickfujita:


nickfujita:
If there are no future planned releases for master before new `dev` upgrade, please just close.

PR #324, opened 2018-10-04
chore(wallet): fixed balance calculation when marking as confirmed

mhuggins:
This fixes an issue where calling `balance.confirm()` doesn't update the balance in the same sense that `balance.applyTx(tx, true)` does.

I'm not sure if you're taking fixes for the v3 branch, so if not, please let me know.


mhuggins:
cc @snowypowers

PR #328, opened 2018-10-17
Add NNS Plugin

nickfujita:
This is a plugin for NEO Name Service. The initial use case is to provide base level support for resolving domain names to addresses.


canesin:
@nickfujita there is currently no plans to support NNS.

WyattMufson:
Even if individual wallets don't want to support NNS, why not add it to neon-js so those that do have the ability? Why leave out features people want?

PR #329, opened 2018-10-22
chore(balance): prevent applyTx from being called twice for contracts & invocations

mhuggins:
This fixes an issue I discovered with transactions being double-applied to balances for "contract" and "invocation" transactions.  When using `api.sendAsset` or `api.doInvoke`, the `balance.applyTx` function was effectively being called twice:

**For sendAsset:**

1. via `createTx` → `createContractTx` → `Transaction.calculate` → `Balance.applyTx`
2. via `sendTx` → `applyTx`

**For doInvoke:**

1. via `createTx` → `createInvocationTx` → `Transaction.calculate` → `Balance.applyTx`
2. via `sendTx` → `applyTx`

This removes the call for `Balance.applyTx` from `Transaction.calculate`, leaving it as part of the core api flow.

### DIFF
--- a/src/transactions/transaction.js
+++ b/src/transactions/transaction.js
@@ -223,7 +223,6 @@ class Transaction {
     const { inputs, change } = core.calculateInputs(balance, this.outputs, this.gas, strategy, fees)
     this.inputs = inputs
     this.outputs = this.outputs.concat(change)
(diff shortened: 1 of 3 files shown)


mhuggins:
If this looks good, I can open a PR with the same change for master.

PR #331, opened 2018-10-23
Feature, generic neon-ns implementation

merl111:


packages/neon-ns/src/abi.ts
```diff
@@ -1,28 +1,30 @@
...
-
-    return sb.emitAppCall(scriptHash, operation, args);
-  };
-}
+import { sc, u } from "@cityofzion/neon-core";
+
+export function resolve(
+    name: string,
+    scriptHash: string = "348387116c4a75e420663277d9c02049907128c7", // default for now nns jump contract
+    operation: string = "resolve"
```
snowypowers: Why do we want to accept a custom operation when the method name is the same?

merl111: > Seems like something went wrong with the branching. There are a lot of my changes that already exist on dev. Please do something about it.

Might be because I pulled the latest dev branch to my local branch first. I can remove those commits.

> I would prefer the name neon-domain to make a clear distinction about the feature and to avoid any confusion if people were to shorten the neon to n.

No problem, I will rename it.

> I think the design here is not clear enough to be generalised. I do not see a way where I am able to choose the implementation that I need.

Will be changed aswell, you are right, there has to be support for other domains too.

> Im not sure about naming it ABI. As much as I like this repository styled in the way I wrote the NEP5 repo, there was a clear template to follow whereas it is arbitrary here. Calling it ABI would be setting the standard ourselves and expecting others to follow. That is not the intent of this.

I just used the already available structure. I will redesign the package and think about a new structure and naming, maybe you already have something in mind?

snowypowers: I would name it core but i think the ABI file will go away once you restructure the package. My initial thought would be to follow the provider design in neon-api where we have a `Provider` interface with neoscan and neonDB implementations. You are free to suggest your own design/ tweaks as long as its achieves the purpose of us being able to select the network, implementation and desired contract.

PR #333, opened 2018-10-26
getVersion bugfix test

phetter:
RPRClient getVersion() was not properly verifying a variable before access. I can't say I see why we need the version regex, but I've modified as-is.

The error that triggered this sequence of events:
```
node neon-js/cli/rpc/getversion.js -n http://seed2.neo.org:20332
(node:13603) UnhandledPromiseRejectionWarning: TypeError: Cannot read property '1' of null
    at /home/fet/nwd/phetter/neotools/node_modules/@cityofzion/neon-js/lib/index.js:36882:63
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
(node:13603) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:13603) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### DIFF
--- a/packages/neon-core/src/rpc/RPCClient.ts
+++ b/packages/neon-core/src/rpc/RPCClient.ts
@@ -239,8 +239,12 @@ export class RPCClient {
   public async getVersion(): Promise<string> {
     try {
       const response = await this.execute(Query.getVersion());
-      const newVersion = response.result.useragent.match(versionRegex)[1];
-      this.version = newVersion;
+      if (response && response.result && response.result.useragent) {
+        const newVersion = response.result.useragent.match(versionRegex)[1];
+        this.version = newVersion;
+      } else {
+        this.version = RPC_VERSION;
+      }
       return this.version;
     } catch (err) {
       if (err.message.includes("Method not found")) {


snowypowers:
The original intent was to block unsupported RPC calls. However, this is very rarely used with how slow RPC moves.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -239,8 +239,12 @@ export class RPCClient {
   public async getVersion(): Promise<string> {
     try {
       const response = await this.execute(Query.getVersion());
-      const newVersion = response.result.useragent.match(versionRegex)[1];
-      this.version = newVersion;
+      if (response && response.result && response.result.useragent) {
+        const newVersion = response.result.useragent.match(versionRegex)[1];
+        this.version = newVersion;
+      } else {
+        this.version = RPC_VERSION;
```
snowypowers: this hides the fact that the response was not to specification. If we are to guard, we should be throwing if `result` or `useragent` is not found.

snowypowers:
@phetter I have dug into this further and I know the reason for it breaking. The return value changed in v2.9.0 and thus broke the regex.

However, I do think the regex was a mistake so here's a good time to make it better. The current standards are:

```
/NEO:2.7.6.1/
/NEO-PYTHON:0.7.7/
/Neo:2.9.0/
/Neo:2.9.0-preview1/
```

As such, we can change the regex to `/\/(.+):(.+)\//` which will accommodate all current version syntax. 

Do let me know if you do not like regex :)

packages/neon-core/src/rpc/RPCClient.ts
phetter: I agree. I will rework and do some more testing. I'm glad we were able to start the discussion and thanks for the help.

phetter:
I'm fine with regex. :+1: I'll review and circle back.

phetter:
I opened a new PR. This one is defunct.

PR #336, opened 2018-10-31
Update createscript.md

jhwinter:
I've been trying to use neon-js to develop a dApp and noticed that my script was failing due to my passing in the parameters improperly.


snowypowers:
👍 Will deploy when `dev` gets merged into master

PR #337, opened 2018-11-02
feature, neon-domain, generic approach for a multi provider resolve interface

merl111:
From last PR:

>Seems like something went wrong with the branching. There are a lot of my changes that already exist on dev. Please do something about it.

removed commits. Should be good now.

>I would prefer the name neon-domain to make a clear distinction about the feature and to avoid any confusion if people were to shorten the neon to n.

renamed to neon-domain

> I think the design here is not clear enough to be generalised. I do not see a way where I am able to choose the implementation that I need.
>Let's say tomorrow, I want to publish my own contract resolving my own domains whatever.snowy. I can't use the ABI or the genericResolve because they are hardcoded to .neo. How do I structure the code such that this plugin is able to accept PRs easily from other implementations?

The plugin can now be used with different contracts and resolveDomain supports whatever tld one would like to resolve.

>Im not sure about naming it ABI. As much as I like this repository styled in the way I wrote the NEP5 repo, there was a clear template to follow whereas it is arbitrary here. Calling it ABI would be setting the standard ourselves and expecting others to follow. That is not the intent of this.

ABI was removed completely after restructuring the package, similar to neon-api now.


packages/neon-domain/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "peerDependencies": {
+    "@cityofzion/neon-core": "^4.0.0"
+  },
+  "files": [
+    "lib/"
+  ],
+  "dependencies": {
+    "jest": "^23.6.0",
+    "jest-cli": "^23.6.0"
```
snowypowers: jest should not be in dependencies.

packages/neon-domain/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "access": "public"
+  },
+  "keywords": [
+    "neo",
+    "antshares",
+    "javascript",
+    "libraries",
+    "neon-ns"
+  ],
+  "author": "Mathias Enzensberger <merl111@protonmail.com> (https://github.com/merl111)",
```
snowypowers: If you have used nick's code in any way, please add him as a contributor.

packages/neon-domain/src/provider/NeoNS/class.ts
```diff
@@ -0,0 +1,26 @@
+import { logging } from "@cityofzion/neon-core";
+import { DomainProvider } from "../common";
+import {
+    resolveDomain 
+} from "./core";
+
+const log = logging.default("neon-domain");
+
+export class NeonNS implements DomainProvider {
```
snowypowers: NeoNS, not NeonNS.

packages/neon-domain/src/provider/common.ts
```diff
@@ -0,0 +1,27 @@
+import { logging, rpc, sc, u } from "@cityofzion/neon-core";
+
+export interface DomainProvider {
+  name: string;
+  resolveDomain(url: string, domain: string, tld: string): Promise<string>;
+}
+const log = logging.default("neon-domain");
+
+export async function resolve(
```
snowypowers: I dont think we can define this as common functionality. Let's move it back under the nns domain.

packages/neon-domain/src/provider/NeoNS/core.ts
```diff
@@ -0,0 +1,49 @@
...
+    value: "addr"
+  };
+
+  const empty = {
+    type: "String",
+    value: ""
+  };
+
+  const tldRegEx = ".".concat(tld).concat("$"); 
+  const regExp = new RegExp(tldRegEx);
```
snowypowers: I would prefer to simplify it:
```
const regExp = new RegExp(`.${tld}$`)
```

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,51 @@
...
+  provider = new neonJs.domain.nns.instance("348387116c4a75e420663277d9c02049907128c7");
+  try {
+    const testUrl = "http://seed1.cryptoholics.cc:10332";
+    const config = {
+      api: {
+          getRPCEndpoint: jest.fn().mockImplementationOnce(() => testUrl)
+      } as any,
+          address: ""
+    }
+      const result = await neonJs.api.fillUrl(config);
```
snowypowers: I don't understand why are we depending on the api plugin for test setup but mocking it out. This doesn't really serve any purpose. 

Do look at the integration test for nep5. I have a simple loop that attempts to get a good rpc node to test again.

packages/neon-domain/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "keywords": [
+    "neo",
+    "antshares",
+    "javascript",
+    "libraries",
+    "neon-ns"
```
snowypowers: remove `neon-ns`

packages/neon-domain/src/provider/common.ts
```diff
@@ -0,0 +1,27 @@
+import { logging, rpc, sc, u } from "@cityofzion/neon-core";
+
+export interface DomainProvider {
+  name: string;
+  resolveDomain(url: string, domain: string, tld: string): Promise<string>;
```
snowypowers: is it possible for us to simplify this down to just `(url, domain)`? From a user perspective, I think many are used to just typing in the entire address as opposed to having separate fields. We can do the check for tld internally. 

This also seems like an implementation detail that is provider specific.

packages/neon-domain/README.md
```diff
@@ -0,0 +1,25 @@
...
+## Installation
+
+```sh
+yarn i @cityofzion/neon-domain
+```
+
+```js
+const neonCore = require("@cityofzion/neon-core");
+const apiPlugin = require("@cityofzion/neon-api");
+const doaminPlugin = require("@cityofzion/neon-api");
```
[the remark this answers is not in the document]
snowypowers: Remove apiPlugin detail if we are not using it at all.

packages/neon-domain/src/provider/NeoNS/core.ts
```diff
@@ -0,0 +1,49 @@
...
+} from "../common";
+
+const log = logging.default("neon-domain");
+
+const operation = "resolve";
+
+/**
+ * Get balances of NEO and GAS for an address
+ * @param url - URL of a neonDB service.
+ * @param address - Address to check.
```
mhuggins: These don't match the params in the function below.

packages/neon-domain/src/provider/NeoNS/core.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @return  Balance of address
+ */
+export async function resolveDomain(
+  url: string,
+  contract: string,
+  domain: string,
+  tld: string
```
mhuggins: The `domain` and `tld` args both seem like unnecessary duplicate pieces of data.  They are both things that can be derived from `url`.

packages/neon-domain/src/provider/NeoNS/core.ts
```diff
@@ -0,0 +1,49 @@
...
+import {
+   resolve,
+} from "../common";
+
+const log = logging.default("neon-domain");
+
+const operation = "resolve";
+
+/**
+ * Get balances of NEO and GAS for an address
```
mhuggins: This doesn't match what the function is doing.

packages/neon-domain/src/provider/NeoNS/core.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @param url - URL of a neonDB service.
+ * @param address - Address to check.
+ * @return  Balance of address
```
mhuggins: This doesn't match what the function is returning.

packages/neon-domain/src/provider/NeoNS/class.ts
```diff
@@ -0,0 +1,26 @@
...
+  public get name() {
+      return `NeonNs[${this.contract}]`;
+  }
+
+  constructor(contract: string) {
+    this.contract = contract;
+    log.info(`Created NeonNS DomainProvider: ${this.contract}`);
+  }
+
+  public resolveDomain(url: string, domain: string, tld: string): Promise<string> {
```
mhuggins: This doesn't match the arguments you provided in `core.ts`.

merl111:
@mhuggins @snowypowers I implemented all the requested changes, please review again.

merl111:
> Naming is all over the place: NeoNS, NeonNS, nns. Choose one and stick with it. I would prefer nns since that is what they are mostly know by.
Sorry I forgot to rename everything to nns, if you still want to name it nns, I will do it.

>I see a bin file. Is it something we need?
Can you tell me which one? I am sure we don't need it but I checked my commits and I couldn't find any binary.

>Remove commented code if we do not need it. Remove neon-api if not required.
Done

>Edit the tsconfig at the root to include this package.
Done

snowypowers:
`packages/neon-domain/src/provider/NeoNS/.responses.ts.swp` is a bin file that was included.

merl111:
> `packages/neon-domain/src/provider/NeoNS/.responses.ts.swp` is a bin file that was included.

Oh that's an invisible VI swap file committed by accident. Removed it.

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,59 @@
+import apiPlugin from "@cityofzion/neon-api";
+import * as neonCore from "@cityofzion/neon-core";
+import domainPlugin from "../src/index";
+
+const neonWithDomain = apiPlugin(neonCore);
```
snowypowers: We can remove the apiPlugin dependency

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,59 @@
...
+const neonJs = domainPlugin(neonWithDomain);
+const { CONST, rpc, u, api, wallet } = neonJs;
+
+let MAINNET_URL = "";
+let provider;
+
+const MAINNET_URLS = [
+  "http://seed1.cryptoholics.cc:10332",
```
snowypowers: please rely on official nodes or coz nodes only

merl111:
>Slight test cleanup

Removed apiPlugin and unofficial RPC node from test.

snowypowers:
You have not addressed the comment about the documentation that doesnt match the function.

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,56 @@
+import * as neonCore from "@cityofzion/neon-core";
+import domainPlugin from "../src/index";
+
+const neonJs = domainPlugin(neonCore);
+const { CONST, rpc, u, api, wallet } = neonJs;
+
+let MAINNET_URL = "";
+let provider;
+
+const MAINNET_URLS = [
```
metachris: Isn't there a full list of mainnet nodes already somewhere inside? this list is very short

merl111:
@snowypowers I missed that comment, fixed now.
@metachris I couldn't find a list, so I created one with all NGD and CoZ nodes.

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,59 @@
...
+});
+
+describe("domainResolve", () => {
+  test("name found", async () => {
+    const address = await provider.resolveDomain(
+      MAINNET_URL,
+      "test.neo",
+      "neo"
+    );
+    expect(wallet.isAddress(address)).toBe(true);
```
mhuggins: The result won't be an address for nOS domains.  Since this is specific to NNS, perhaps this file be renamed from "main" to "nns" or "NeoNS".

packages/neon-domain/src/provider/NeoNS/main.ts
```diff
@@ -0,0 +1,25 @@
...
+  operation: string, 
+  args: any[]
+) {
+  const sb = new sc.ScriptBuilder();
+  const script = sb.emitAppCall(scriptHash, operation, args).str;
+  try {
+    const res = await rpc.Query.invokeScript(script).execute(url);
+    return rpc.StringParser(res.result.stack[0]);
+  } catch (err) {
+    log.error(`resolveDomain failed with : ${err.message}`);
```
mhuggins: `resolveDomain` => `resolve`?

mhuggins: Is there any benefit to this try/catch?  It looks like the only benefit is adding the function name that failed, but the error's stack trace will already provide that.

packages/neon-domain/__integration__/main.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  test("name not found", async () => {
+    try {
+      const address = await resolveDomain(
+        MAINNET_URL,
+        "alkdjfklasjdlfkjasdklf.neo",
+        "neo"
+      );
+    } catch (err) {
+      expect(true).toEqual(true);
```
mhuggins: It looks like this test is not testing what is intended.  You'r calling `resolveDomain` here, which I think should be `provider.resolveDomain`.  The error is being caught due to a typo rather than the the name not being found.  A better check here would be something like:

```js
expect(err.message).toMatch(/not found/)
```

packages/neon-domain/__integration__/main.ts
merl111: Ok, renamed it.

packages/neon-domain/src/provider/NeoNS/main.ts
merl111: removed resolve and put those three lines to resolveDomain. No the try/catch was not really necessary, removed it too.

packages/neon-domain/__integration__/main.ts
merl111: Testing against an empty string now, which is returned if no address was found and fixed the function call.

packages/neon-domain/__integration__/urls.ts
```diff
@@ -0,0 +1,26 @@
+
+export function getMainnetUrls(): string[] {
```
snowypowers: This file is failing the integration tests because it tries to find a test inside. Let's just add this function into `__integration__/NeonNS.ts` and delete this file.

merl111: fixed, but maybe there is a way to have the urls.ts somewhere else to use it for all tests?

snowypowers: yeah that can be an improvement in a separate PR.

snowypowers:
Its been a long review. Thanks for the contribution!

PR #339, opened 2018-11-05
Chore(rpc client): fix getVersion 

phetter:
getVersion wasn't properly handling version responses. 

I dropped the regex and went with a string split approach instead. 

I tested against test net and it is reporting versions properly.

### DIFF
--- a/packages/neon-core/src/rpc/RPCClient.ts
+++ b/packages/neon-core/src/rpc/RPCClient.ts
@@ -236,21 +236,33 @@ export class RPCClient {
   /**
    * Gets the version of the NEO node. This method will never be blocked by version. This method will also update the current Client's version to the one received.
    */
-  public async getVersion(): Promise<string> {
-    try {
-      const response = await this.execute(Query.getVersion());
-      const newVersion = response.result.useragent.match(versionRegex)[1];
-      this.version = newVersion;
-      return this.version;
-    } catch (err) {
-      if (err.message.includes("Method not found")) {
-        this.version = RPC_VERSION;
-        return this.version;
-      } else {
-        throw err;
-      }
-    }
-  }
+   getVersion() {
+       return __awaiter(this, void 0, void 0, function* () {
+           try {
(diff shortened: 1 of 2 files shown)


phetter:
I just realized I think I PR'ed off the wrong source. There is an awaiter function in this.

phetter:
I think that's it.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -238,15 +238,25 @@ export class RPCClient {
...
     try {
-      const response = await this.execute(Query.getVersion());
-      const newVersion = response.result.useragent.match(versionRegex)[1];
-      this.version = newVersion;
+      const response = yield this.execute(Query_1.default.getVersion());
+      if (response && response.result && response.result.useragent) {
+        var version = response.result.useragent.split(':');
+        version[0] = version[0].substring(1);
+        version[1] = version[1].substring(0, version[1].length-1);
+        const newVersion = version[1];
```
snowypowers: Instead of stripping it seperately, lets start by stripping the leading and trailing `/`. Then we can use array destructuring to clarify our intent instead of using array indices:

```
const responseLength = response.result.useragent.length;
const strippedResponse = response.result.useragent.substring(1, responseLength-1);
const [header, newVersion, ..._] = strippedResponse.split(':');
this.version = newVersion;
```

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -238,15 +238,25 @@ export class RPCClient {
    */
   public async getVersion(): Promise<string> {
     try {
-      const response = await this.execute(Query.getVersion());
-      const newVersion = response.result.useragent.match(versionRegex)[1];
-      this.version = newVersion;
+      const response = yield this.execute(Query_1.default.getVersion());
```
snowypowers: I see `yield`. I think something went wrong here.

phetter:
getVersion test has been updated with passing status locally.

PR #343, opened 2018-11-11
feat(neon-nep5): added getTokens functionality

DalderupMaurice:
You're able to get multiple token balances in 1 call, but not the ability to get multiple tokens (with more detailed info) in 1 call.
This PR resolves this issue, with the addition of `getTokens`

### DIFF
--- a/packages/neon-nep5/__integration__/main.ts
+++ b/packages/neon-nep5/__integration__/main.ts
@@ -107,13 +107,8 @@ describe("getTokens", () => {
   test("without balance", async () => {
     const info = await getTokens(TESTNET_URL, [
       CONST.CONTRACTS.TEST_RPX,
-      // CONST.CONTRACTS.TEST_RHTT4,
       CONST.CONTRACTS.TEST_LWTF,
       CONST.CONTRACTS.TEST_NXT
-      // CONST.CONTRACTS.TEST_RPX,
-      // CONST.CONTRACTS.TEST_RHTT4,
-      // CONST.CONTRACTS.TEST_LWTF,
-      // CONST.CONTRACTS.TEST_NXT
     ]);
 
     expect(Object.keys(info).map(key => info[key])).toEqual(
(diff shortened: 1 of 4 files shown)


packages/neon-nep5/src/main.ts
```diff
@@ -141,3 +141,78 @@ export async function getToken(
     throw err;
   }
 }
+
+/**
+ * Retrieves the complete information about all (or a list of) tokens.
```
snowypowers: `all (or a list of)` suggests that the scriptHashArray has two interpretations but I don't see that. I think it will be good to include in the docstring the properties returned.

packages/neon-nep5/src/plugin.ts
```diff
@@ -1,4 +1,4 @@
 import * as abi from "./abi";
-import { getToken, getTokenBalance } from "./main";
+import { getToken, getTokenBalance, getTokens } from "./main";
 
-export { abi, getToken, getTokenBalance };
+export { abi, getToken, getTokens, getTokenBalance };
```
snowypowers: I apologise for my error but since you are touching this file, could you also include the export for `getTokenBalances`? I did not realise that was not exported.

packages/neon-nep5/src/main.ts
```diff
@@ -141,3 +141,78 @@ export async function getToken(
...
+      };
+      
+      if(!obj.balance) delete obj.balance;
+
+      result.push(obj);
+    }
+    return result;
+  } catch (err) {
+    log.error(`getToken failed with : ${err.message}`);
+    throw err;
```
snowypowers: I think it is weird that we only try catch the invocation and parsing part but not the script construction when you are going to log an generic error.

Either:

1. Try catch the entire thing to log the generic error.
2. Split the try catch into logical components, so one trycatch for scriptbuilder, one for invocation and one for parsing. 
3. Don't try catch at all since the `err.message` is likely to bubble all the way up and we are not adding any good contextual information to it here.

snowypowers: I know this is likely because of the structure of `getTokenBalances` but I would like to take this chance to revisit error handling. Might have to invest some time into making it consistent throughout the codebase. I think you, as a dev working with it, is in a good spot to provide some feedback on what would improve your experience :)

packages/neon-nep5/__integration__/main.ts
```diff
@@ -97,3 +102,62 @@ describe("getToken", () => {
...
+          decimals: 8,
+          name: "LOCALTOKEN",
+          symbol: "LWTF",
+          totalSupply: 2270000
+        },
+        {
+          decimals: 8,
+          name: "NEX Template",
+          symbol: "NXT",
+          totalSupply: 2508560
```
snowypowers: Im not certain that these supply numbers will stay static. Do you know? I will accept a test just checking for number type and maybe larger than a arbitrary number (the best we can do if these numbers flutter but I doubt it will go down).

packages/neon-nep5/src/plugin.ts
DalderupMaurice: Done!

packages/neon-nep5/src/main.ts
DalderupMaurice: Modified the text as that was unintended :P

packages/neon-nep5/src/main.ts
DalderupMaurice: In my opinion, this is not the place where the "truly useful" are created.
If anything goes wrong on a deeper level (at the creation of the build script or something), this is the place where it should provide the necessary useful message(s) for the user to understand what went wrong. 

I think it's sufficient to have the underlying error (if the message is clear enough and if it's useful for the user) with the called function name contained in the error as done right here --> `log.error("getToken failed with : ${err.message}");`

Although this is just my sole opinion and better ways are 100% certainly possible. Maybe this is something to discuss on Slack/Discord instead.

As a last note, I've now move the try/catch up. If you'd like it to be reverted right now or applied to the other functions, I'm willing to do so!

packages/neon-nep5/__integration__/main.ts
DalderupMaurice: Apparently, it is possible for the Total Supply to decrease. I've adjusted the tests

packages/neon-nep5/src/main.ts
phetter: For debugging, I'm a big fan of grouping error messages by logical abstractions with verbosity to make it quick and easy to find a chunk of code that is problematic. It should be left up to project API implementations to make the errors user-friendly

DalderupMaurice: I just noticed that `getTokenBalances` has a try-catch in the loop and outside of it. Would you like me to nest it like that aswell?

PR #347, opened 2018-11-14
fix(neon-core): Fix ContractParam.Integer transformation

### LINKED ISSUE #345: Wrong BigInteger conversion
It seems that there is an error in the BigInteger conversion, the following example shows the issue.
```js
function1 = async (address, label, value, balance) => {
    const script = this.buildScript('run', [
      string('function'),
      byteArray(address, 'address'),
      string(label),
      integer(value)
    ]);

    return this.invoke({ script, balance });
  };
```
When I run function1 with these parameters; `AddressXXXX...`, `'label'`, `14256661` it generates the following script code.
```
PUSHBYTES3: 0x158ad9 --> 14256661 (wrong)
PUSHBYTES5: 0x6c6162656c --> label
PUSHBYTES20: 0xXXXXX... --> AddressXXXX...
PUSHBYTES8: 0x66756e6374696f6e --> function
PUSH4 PACK
PUSHBYTES3: 0x72756e --> run
APPCALL: xxxx... --> contract_address
```

The problem is `14256661` has been converted to `0x158ad9` what is not `14256661` , is -`2520555` in BigInteger and since VM interprets values as BigInteger, the conversion is wrong.
The right value for `14256661` is `158ad900`

Here a C# example of the right conversions for `14256661` https://dotnetfiddle.net/IfHgvp

snowypowers:
C# BigInteger uses two's complement which wasn't reflected by the `parseInt` of javascript. This solution brings in Igor's `csbiginteger` package which is a javascript implementation of the c# BigInteger class. With this, we should be able to convert correctly. It's a good place to start testing the package out too.

Fix #345

### DIFF
--- a/packages/neon-core/__tests__/sc/ScriptBuilder.ts
+++ b/packages/neon-core/__tests__/sc/ScriptBuilder.ts
@@ -1,3 +1,4 @@
+import ContractParam from "../../src/sc/ContractParam";
 import OpCode from "../../src/sc/OpCode";
 import ScriptBuilder from "../../src/sc/ScriptBuilder";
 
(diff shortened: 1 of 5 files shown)


mhuggins:
🙏

PR #358, opened 2018-11-20
fix retrieveAppCall for toScriptParams

Sotatek-NhanPhung:
If input Script of Invoke Transaction is array, method toScriptParams will throw an error, it wrong
**Example** 
  `txid` : 69e8341aa13f95a1eef43ba322e298bcbe354d5659b0f45fdffc47111fb9f2e1
  `network:` Testnet


snowypowers:
Do we have an example? Please add a test with the data.

Sotatek-NhanPhung:
```

  const invoTransaction ='69e8341aa13f95a1eef43ba322e298bcbe354d5659b0f45fdffc47111fb9f2e1';
  const endpoint = 'http://seed4.neo.org:20332';
  const rawTransaction = ( 
await  rpc.Query.getRawTransaction(invoTransaction).execute(endpoint)).result;
let inputScript  = rawTransaction.script;
    if(!Neon.u.isHex){
      throw new Error('input script is not hex type')
    }

    let script = new sc.ScriptBuilder(inputScript).toScriptParams();
    let scriptParams = [];
    
    if(Array.isArray(script)){

      script.forEach(e => {

        let method = Neon.u.hexstring2str(e.args[0]);

        let tokenInfo = {
          value : Neon.u.fixed82num(e.args[1][0]),
          to : this.ScriptHashToAddress(e.args[1][1]),
          from : this.ScriptHashToAddress(e.args[1][2])
        }
        scriptParams.push({method:method,scriptHas:e.scriptHash, tokenInfo:tokenInfo})
      })
    }

    else if(typeof script === 'object'){
      let method = Neon.u.hexstring2str(script.args[0]);

        let tokenInfo = {
          value : Neon.u.fixed82num(script.args[1][0]),
          to : this.ScriptHashToAddress(script.args[1][1]),
          from : this.ScriptHashToAddress(script.args[1][2])
        }

        scriptParams.push({method:method,scriptHas:e.scriptHash, tokenInfo:tokenInfo})
    }

    else{
      throw new Error('input script is not valid')
    }
    
    return scriptParams
  }
  ScriptHashToAddress(hex) {
    return wallet.getAddressFromScriptHash(Neon.u.reverseHex(hex));
  
```
this transaction had accepted in the test network, but  `new sc.ScriptBuilder(inputScript).toScriptParams()` will throw error `"No Scripthash found!"`. i search in ScriptBuilder and found this :
```
function retrieveAppCall(sb) {
    const output = {
        scriptHash: "",
        args: []
    };
    while (!sb.isEmpty()) {
        const b = sb.read();
        const n = parseInt(b, 16);
        switch (true) {
            case n === 0:
                output.args.unshift(0);
                break;
            case n < 75:
                output.args.unshift(sb.read(n));
                break;
            case n >= 81 && n <= 96:
                output.args.unshift(n - 80);
                break;
            case n === 193:
                const len = output.args.shift();
                const cache = [];
                for (let i = 0; i < len; i++) {
                    cache.unshift(output.args.shift());
                }
                output.args.unshift(cache);
                break;
            case n === 102:
                sb.pter = sb.str.length;
                return output;
                break;
            case n === 103:
                output.scriptHash = u_1.reverseHex(sb.read(20));
                output.useTailCall = false;
                return output;
            case n === 105:
                output.scriptHash = u_1.reverseHex(sb.read(20));
                output.useTailCall = true;
                return output;
            case n === 241:
                break;
            default:
                throw new Error(`Encounter unknown byte: ${b}`);
        }
    }
    if (output.scriptHash === "") {
        throw new Error("No Scripthash found!");
    }
    return output;
}
```
 n = 102 when ScriptBuilder read to (script.length -16), i think this word to end of array , but switch case not return output , it is break , and run to ` 
if (output.scriptHash === "") {
        throw new Error("No Scripthash found!");
    } `
    and throw error , this is problem!
Please , check it

snowypowers:
Ah I see the issue. There are dangling instructions at the end of the script which is interfering with the processing. Yes, I think the short fix here is to actually remove the `throw No ScriptHash found` at the end as we should not reach that part if we actually find a APPCALL or TAILCALL. Im thinking that we return `null` at the bottom so this way, all the redundant instructions are thrown away and we only have the appcalls.

Chubby-Chocobo:
I clean the commit and add unit test for it.
Please help to take a look and merge if it's ok @snowypowers 
Thank you.

PR #363, opened 2018-11-26
Fix variable name

corollari:
'neon-js' is illegal as a variable name in JavaScript because dashes ('-') are not allowed in variable names. This results in an error when executing the script.
This PR fixes the variable name by replacing the dash with an underscore.


snowypowers:
Hi I would prefer to use `neonJs` over underscores to keep everything else consistent. Most of the codebase is running camelCase.

Thanks for noticing and contributing!

corollari:
Okay, I'll update the PR

PR #370, opened 2018-12-04
Updating dependency to csbiginteger 3.0.0

igormcoelho:
Snowy, now project was moved to NeoResearch page for future and more stable maintainance. I'm using "bn.js" library to handle internals now, so it's very stable. The idea is to manage it to be compatible with C# representation, but at least now the code is much shorter and easier to maintain. I believe this major can last a long time now, since it's very tested on many cases.

### DIFF
--- a/packages/neon-core/package.json
+++ b/packages/neon-core/package.json
@@ -46,7 +46,7 @@
     "bs58": "^4.0.1",
     "bs58check": "^2.1.2",
     "crypto-js": "^3.1.9-1",
-    "csbiginteger": "^2.12.0",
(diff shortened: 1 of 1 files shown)


igormcoelho:
Ok, no worry, I thought that was being used in the base already. Anyway, it's quite similar to bn, with few exceptions... for you it may not matter much, but since I want to build a neovm simulator, I need it exactly the same, thats the goal of the project. Tske your time.

ixje:
Obsolete as `neon-core` no longer relies on this package (not for 2.x and not for 3.x)

PR #375, opened 2018-12-06
Adding Construct for GAS claiming

vncoelho:


snowypowers:
1. I am moving towards a plan where I generate the markdown files from the js files found in examples. So the correct place to edit this should be in the js file.

2. The scripts are intended to be runnable without much editing. Adding this might cause confusion and it is not contributing to the purpose of the script. I suggest adding another script or constructing this as a comment instead.

PR #379, opened 2018-12-17
feat(ledger): Implement neon-ledger

snowypowers:
This is a wrapper for interacting with ledgerhq/hw-transport.

Due to some issues with the current ledger package, I will not merge this until they resolve it.

TLDR: They removed `babel-transform-runtime` which led to their production code relying on the `babel-runtime` package. You will need to require the babel runtime package before importing the ledger packages.


DalderupMaurice:
https://github.com/LedgerHQ/ledgerjs/releases/tag/v5.0.0

The 5.0 release of ledger includes a rework of their babel configuration. Is this the issue you were waiting for?

PR #403, opened 2019-04-13
Chore: Rewrite Imports for Tests to get all Integration and Unit Tests to Pass on Linux

phetter:
This needs a review. Any help, input, or feedback would be appreciated. 

This isn't the first time I've had this problem. I wasted a bunch of time trying to resolve this last year, but I never committed myself to resolve it because I thought I was doing something wrong. It never got closure. This time I decided to make it work.

So I spent 3 days going around and around with this. The only fix I could find was to rewrite all the imports to use a local path instead of the @cityofzion/neon-core distribution link in each of the test files.

The builds would work but the tests wouldn't if I followed the readme. Performing ```npm i @cityofzion/neon-js``` as described by the readme removes symlinks back into the packages and causes tests to fail. In some cases it also causes conflicts with yarn via jest, lerna, and tslint. It depends on the sequence of commands.

To recapitulate:

I rewrote the imports so the tests work against the code and not the distribution.

Following the instructions on the readme of the repo would build the project, but I couldn't get all of the tests to pass (integration nor unit) unless I rewrote the imports as shown in the files in the PR.

System: 
Ubuntu 18.04
node 8.15.1
npm 6.4.1
yarn 1.15.2

Thanks!


snowypowers:
i dont think platform is an issue here. Have you tried building everything first before running the integration tests? lerna bootstraps it so that when you `import * from '@cityofzion/neon-core' it actually imports from `./packages/neon-core`. So you have to build using `yarn dist` first.

phetter:
Yes, I built everything first. I followed the instructions in the readme starting at the top and went down.

I didn't yarn dist first. It wasn't in the readme. I tried it later during troubleshooting on my own and while investigating other scripts in the package.json as yarn dist is not mentioned in the readme.

I am deeply saddened that I wasted so much time on this just because yarn dist wasn't in the readme, but I'm thankful that you took the time to help me work around this. I suppose this PR can be closed.

The tests passed using the sequence:

npm i @cityofzion/neon-js
yarn
yarn bootstrap
yarn build
yarn lint
yarn build
yarn dist
yarn test:unit
yarn test:integration

I understand the steps above aren't all required, but that is what I ran to see it build and have all tests pass.

Thanks again.

Cheers

Issue #401, opened 2019-04-08
getRPCEndpoint ignores httpsOnly setting


hal0x2328:
The `httpsOnly` member of `settings` doesn't seem to have any impact when selecting a node for RPC requests. Even when set to `true`, http-only nodes are not filtered out. The filtering method tests work properly but in practice it seems like the setting of `httpsOnly` in the global settings doesn't propagate to the `internalSettings` object in https://github.com/CityOfZion/neon-js/blame/master/packages/neon-api/src/provider/neoscan/core.ts so the filter is never executed.

Here's a JSFiddle demonstrating the issue (it's possible it might randomly select an https node on its own, so run it a few times if it doesn't select an http node immediately)

https://jsfiddle.net/s5tz4hg2/

PR #405, opened 2019-04-13
Chore: Make Neoscan getRPCEndpoint Respect httpsOnly Flag

phetter:
This is an integration test and a code change to address the following issue:

https://github.com/CityOfZion/neon-js/issues/401

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ yarn build
 ```sh
 yarn lint
 yarn build
+yarn dist
 yarn test:unit
 yarn test:integration
 ```
--- a/packages/neon-api/__integration__/provider/neoscan-testnet.ts
+++ b/packages/neon-api/__integration__/provider/neoscan-testnet.ts
@@ -1,0 +1,87 @@
+import * as neonCore from "@cityofzion/neon-core";
+
+import apiPlugin from "../../src/index";
+
+const neonJs = apiPlugin(neonCore);
+
+import * as util from "util"
+
+const { rpc, wallet, u } = neonJs;
+
+const net = "TestNet";
+const addr = "APJnYRic9fYo1bLAdrqDbwrsnqoUD92dWo";
+neonJs.settings.networks[net] = new rpc.Network({
+  Name: "TestNet",
+  ExtraConfiguration: {
+    neoscan: "https://neoscan-testnet.io/api/test_net"
+  }
+});
+
+const provider = new neonJs.api.neoscan.instance("TestNet");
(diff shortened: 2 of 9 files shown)


phetter:
The correct way to call this now is:
```
const provider = new neonJs.api.neoscan.instance("TestNet");
provider.set({httpsOnly: true})
const result = await provider.getRPCEndpoint();
```

hal0x2328:
I can confirm after building this branch and using the new settings method it fixes the issue in #401 and I can now create working neon-js examples in jsfiddle.net. Good work!

phetter:
By the looks of this test I'm guessing DNS is down or lagged as I didn't touch this file.

phetter:
I tested and this is not exposed on the neonCore or NeonJs api plugin objects. Am I calling it as you intended? Please see attached.

Screenshot from 2019-04-19 16-24-11

snowypowers:
my fix retains the old way of using it so `neonJs.settings.httpsOnly = true`.

phetter:
I see that it should, but the object isn't updating or exposed as expected. I've rebuilt and dumped the object and I'm not seeing it on the tree. The code also is not responding to toggling the flag.
Screenshot from 2019-04-19 17-25-21

phetter:
I'm pretty sure all we'd need to do is implement again the change I made to packages/neon-core/src/settings.ts to expose the variable on the api. I'm not sure why this was removed in the update you made. Was there a reason you removed this @snowypowers or am I missing something? Thank you!

snowypowers:
```typescript
test("httpsOnly getRPCEndpoint", async () => {
    neonJs.logging.setAll(0); // This enables noisy console logging
    // expect(neonJs.settings.httpsOnly).toBe(false);
    // neonJs.settings.httpsOnly = true
    // expect(neonJs.settings.httpsOnly).toBe(true);
    const result = await provider.getRPCEndpoint();
    expect(typeof result === "string").toBeTruthy();
    expect(result).toEqual(expect.stringContaining("https://"));
  });
```

Look at the results of this test and the results of it with the httpsOnly line uncommented. You should see that the test will not attempt a `getCount` against nonhttps endpoints if the setting is true.

Do change the expect to refer to neonJs instead of neonCore.

phetter:
Awesome, looks good. Thank you for all the help on this.

hal0x2328:
300 points phetter
200 points snowypowers for review and changes

lllwvlvwlll:

PR #412, opened 2019-04-23
feat(core): Implement parsing for issue transaction

### LINKED ISSUE #411: Implement parsing for issue transaction

dauTT:
Closes #411

### DIFF
--- a/packages/neon-core/__tests__/tx/transaction/issueTx.json
+++ b/packages/neon-core/__tests__/tx/transaction/issueTx.json
@@ -1,0 +1,58 @@
+{
+  "1": {
+    "txid": "3631f66024ca6f5b033d7e0809eb993443374830025af904fb51b0334f127cda",
+    "serialized": "01000000019b7cffdaa674beae0f930ebe6085af9093e5fe56b34a5c220ccdcf6efc336fc50000c16ff28623005fa99d93303775fe50ca119c327759313eccfa1c01000151",
+    "deserialized": {
+      "type": 1,
+      "version": 0,
+      "attributes": [],
+      "inputs": [],
+      "outputs": [
+        {
+          "assetId": "c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
+          "value": 100000000,
+          "scriptHash": "1cfacc3e315977329c11ca50fe753730939da95f"
+        }
+      ],
+      "scripts": [
+        {
+          "invocationScript": "",
+          "verificationScript": "51"
+        }
+      ]
+    }
+  },
+  "2": {
+    "txid": "2f3d6d3b3b650a579633954ae1a566b31adfeaa4d7d6c2fc25e7a13c444ee86c",
+    "serialized": "0100000115501d9597417c8f6a4c53533a7faa983fef2a729c63c66fcc41ae3050780e9d000002c1c1ef12fa6af1eb4dd3429222ca5e621d831c4513d77309c9ece6681651de070080faca73f91f008e5beb4320664b7963b815ff358ed83ab1203004e72d286979ee6cb1b7e65dfddfb2e384100b8d148e7758de42e4168b71792c6000743ba40b0000008e5beb4320664b7963b815ff358ed83ab1203004014140d53f183e22d2e6572fb5265815a24d795f9bec351e07f2ac04fc5e04024869fae92a58b3ca230498bed781638acb56c849b686493e3b90fecb925d33197db6a6232102ec4df037b63756e1c5be04a074d61e2bef0e5439917e80461b22ae28cf65c39eac",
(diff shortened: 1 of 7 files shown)


packages/neon-core/src/tx/transaction/IssueTransaction.ts
```diff
@@ -0,0 +1,51 @@
...
+    ss: StringStream,
+    tx: Partial<TransactionLike>
+  ): Partial<TransactionLike> {
+    return {};
+  }
+
+  public readonly type: TransactionType = TransactionType.IssueTransaction;
+
+  public constructor(obj: Partial<TransactionLike> = {}) {
+    super(obj);
```
snowypowers: looks like the latest version is 1 so we should default to setting the version to 1.

dauTT: @snowypowers , are you sure? Which is the latest version that you are looking at?

snowypowers: im looking at neo source code. There is indication of the existance of version 1 and I would assume that we should be using the latest version template if possible. That said, I believe this is a deprecated transaction type so might not be relevant.

dauTT: Actually the issue transaction are not deprecated. If I look at the [C# neo code](https://github.com/neo-project/neo/blob/master/neo/Network/P2P/Payloads/IssueTransaction.cs#L30) form the master branch I wouldn't be sure that the default version should be set to one: 
```C#
 protected override void DeserializeExclusiveData(BinaryReader reader)
        {
            if (Version > 1) throw new FormatException();
        }
```
The valid versions are either 0 or 1. But so far I have found only about 10 issue transactions and all of them have version 0.  I am currently dumping deserialize json transactions form the python bootstrap Mainet chain into mongo db. From there I will be able to make quick queries but it will take a lot of time to dump everything. Do you still want me to set the default version to one or shall we analyse further?

snowypowers: Lets see if there is anything with version 1. As far as I know, its not marked deprecated but in reality it is not used because it is used together with the one of the deprecated ones.

dauTT: The dumping process has covered so far about 10'500'000 transactions. Grouping the transaction by type we get to the following statistics:
```
{ "_id" : "PublishTransaction", "count" : 1 }
{ "_id" : "IssueTransaction", "count" : 6 }
{ "_id" : "EnrollmentTransaction", "count" : 6 }
{ "_id" : "RegisterTransaction", "count" : 13 }
{ "_id" : "MinerTransaction", "count" : 1448857 }
{ "_id" : "ClaimTransaction", "count" : 1914964 }
{ "_id" : "InvocationTransaction", "count" : 3427266 }
{ "_id" : "StateTransaction", "count" : 6 }
{ "_id" : "ContractTransaction", "count" : 3608881 }
```
So far all the issue transactions have version =0. I will wait that the dumping process will stop hoefully in 2-3 days and cross check again whether we have any issue transaction with version 1.

dauTT: After analyzing more 19'000'000  transactions we found only 16 `IssueTransaction` and all of them have version 0. Considering that mainet has about 22'000'000, I would says that it is ok to set the default version of the IssueTransaction to 0. Did you agree?

packages/neon-core/__tests__/tx/transaction/issueTx.json
```diff
@@ -0,0 +1,58 @@
+{
+  "1": {
+    "txid": "3631f66024ca6f5b033d7e0809eb993443374830025af904fb51b0334f127cda",
+    "serialized": "01000000019b7cffdaa674beae0f930ebe6085af9093e5fe56b34a5c220ccdcf6efc336fc50000c16ff28623005fa99d93303775fe50ca119c327759313eccfa1c01000151",
+    "deserialized": {
+      "type": 1,
+      "version": 0,
```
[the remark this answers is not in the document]
dauTT: After analyzing more 19'000'000 transactions we found only 16 IssueTransaction and all of them have version 0. Considering that mainet has about 22'000'000, I would says that it is ok to set the default version of the IssueTransaction to 0. Did you agree? (As a side note, some of the comments on this PR are hidden or removed. I am not sure if it is accidental or on purpose.)

packages/neon-core/src/tx/transaction/IssueTransaction.ts
snowypowers: ok v=0.

hal0x2328:
300 points dauTT

lllwvlvwlll:

PR #416, opened 2019-04-29
fix(test): Fix test helper being hung up on first url

snowypowers:
Extracts helper function to pull urls into a separate file. Also improves the algorithm by cutting the list at a random point so we do not hit the same urls forever and using a good enough heuristic of checking the difference in blocks between 2 urls are similar to ensure we have a good enough url.


hal0x2328:
200 points snowypowers

lllwvlvwlll:

PR #420, opened 2019-05-05
fix(core): Fix decoding of negative Fixed8 hex string

### LINKED ISSUE #419: bug in decoding negative fixed8 from hex string
There is a bug in [`fixed8num`](https://github.com/CityOfZion/neon-js/blob/master/packages/neon-core/src/u/convert.ts#L159). For instance from the following method call, I would have expected to see `-0.00000001` instead of  `184467440737.09552`:
```typescript
let amount = fixed82num("ffffffffffffffff");
// actual amount: 184467440737.09552
// expected amount: -0.00000001
```

dauTT:
Closes #419

### DIFF
--- a/packages/neon-core/src/u/Fixed8.ts
+++ b/packages/neon-core/src/u/Fixed8.ts
@@ -1,7 +1,20 @@
 import BN from "bignumber.js";
 import { reverseHex } from "./misc";
+import { const } from '../../../neon-js/src/index';
 
 const DECIMALS = 100000000;
+
+// This is the maximum hex integer 0x7fffffffffffffff (= 9223372036854775807)
+// that can be converted to Fixed8 by dividing by the 10^8.
+const MAX_FIXED8_HEX = new BN(2).pow(63).minus(1)
+
+// This is the minimum hex integer 0x8000000000000000 (= -9223372036854775808)
+// that can be converted to Fixed8 by dividing by the 10^8.
+const MIN_FIXED8_HEX = new BN(2).pow(63).negated()
+
+// Total number of Fixed8 available. This includes negative and positive 
+// Fixed8 numbers.
+const TOTAL_FIXED8_HEX = new  BN(2).pow(64)
 
 /**
  * A fixed point notation used widely in the NEO system for representing decimals.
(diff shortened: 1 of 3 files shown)


packages/neon-core/src/u/Fixed8.ts
```diff
@@ -1,6 +1,19 @@
...
 
+const DECIMALS = 100000000;
+
+// 0x7fffffffffffffff = 9223372036854775807 (maximum hex fixed8)
+const MAX_FIXED8_HEX = new BN(2).pow(63).minus(1);
+export const MAX_FIXED8 = MAX_FIXED8_HEX.div(DECIMALS);
+
+// 0x8000000000000000 = -9223372036854775808 (minimum hex fixed8)
+const MIN_FIXED8_HEX = new BN(2).pow(63).negated();
+export const MIN_FIXED8 = MIN_FIXED8_HEX.div(DECIMALS);
```
snowypowers: How about changing this to a static readonly property Fixed8.MIN_VALUE and Fixed8.MAX_VALUE? That is similar to Number.MIN_VALUE and Number.MAX_VALUE in js.

packages/neon-core/src/u/Fixed8.ts
```diff
@@ -28,9 +55,20 @@ export class Fixed8 extends BN {
...
-      .toString(16);
+    let hexstring = "";
+    const num = this.times(DECIMALS).round(0);
+
+    hexstring = num.isLessThan(0)
+      ? TOTAL_FIXED8_HEX.plus(num).toString(16) // convert num to two complement
+      : num.toString(16);
+
+    if (hexstring.length > 16) {
+      throw new Error(
```
snowypowers: This is interesting. So we are basically blocking the `toHex()` operation if our final result is outside of Fixed8 boundaries. However, if our internal operations exceed the limits but the final result does not overflow, it is fine and the result will be accurate.

If we really wanted to do this, we should be validating that the number is within bounds after every operation (+, - ,*, / ) so this operation is safe. It does not really make sense to throw on an invalid output when the input is valid.

dauTT: I agree with you that it is better to validate the input of the Fixed8 class. I will add check directly in the Constructor. In this way, we will be stricter and the check will be applied to every operation.

packages/neon-core/src/u/Fixed8.ts
```diff
@@ -3,24 +3,29 @@ import { reverseHex } from "./misc";
...
-const TOTAL_FIXED8_HEX = new BN(2).pow(64);
 /**
  * A fixed point notation used widely in the NEO system for representing decimals.
  * It is basically a hexideciaml integer that is divided by the 10^8.
  * Supports up to 8 decimals and is 8 bytes long.
  * @extends BN
  */
 export class Fixed8 extends BN {
+  // The maximum Fixed8 is obtained by dividing 0x7fffffffffffffff (= 9223372036854775807) with 10^8.
+  public static MAX_VALUE = new BN(2)
```
snowypowers: this should return a Fixed8?

packages/neon-core/src/u/Fixed8.ts
```diff
@@ -3,24 +3,29 @@ import { reverseHex } from "./misc";
...
+
+  // The minimum Fixed8 is obtained by dividing 0x8000000000000000 (= -9223372036854775808) with 10^8.
+  public static MIN_VALUE = new BN(2)
+    .pow(63)
+    .negated()
+    .div(DECIMALS);
+
+  // TOTAL represents the total number of Fixed8 available. This inludes positve and negative fixed8 numbers.
+  // 2^64 = 0xffffffffffffffff = 18446744073709551615
+  public static TOTAL = new BN(2).pow(64);
```
snowypowers: not sure if i want to expose this. What is the usecase?

dauTT: You right. I will not expose it.

packages/neon-core/src/u/Fixed8.ts
dauTT: Ok, Got it.

packages/neon-core/src/u/Fixed8.ts
```diff
@@ -1,30 +1,34 @@
 import BN from "bignumber.js";
 import { reverseHex } from "./misc";
+import { const } from '../../../neon-js/src/index';
```
snowypowers: this is the issue.

hal0x2328:
200 points dauTT

lllwvlvwlll:
https://neoscan.io/transaction/589328c622e101065f6717d61a332f93aac71cc68b665e194ef8c874a7e26669

PR #425, opened 2019-05-11
feat(core): Implement parsing for publish transaction

### LINKED ISSUE #424: Implement parsing for publish transaction

dauTT:
Closes #424


packages/neon-core/src/tx/transaction/PublishTransaction.ts
```diff
@@ -0,0 +1,195 @@
...
+  hexstring2str,
+  num2hexstring,
+  num2VarInt,
+  str2hexstring,
+  StringStream
+} from "../../u";
+import { BaseTransaction, TransactionLike } from "./BaseTransaction";
+import TransactionType from "./TransactionType";
+
+export enum ContractParameterType {
```
snowypowers: This is defined at `sc.ContractParamType`.

packages/neon-core/src/tx/transaction/PublishTransaction.ts
```diff
@@ -0,0 +1,195 @@
...
+  needStorage: boolean;
+  name: string;
+  codeVersion: string;
+  author: string;
+  email: string;
+  description: string;
+}
+
+export class PublishTransaction extends BaseTransaction {
+  public static version: number = 0;
```
snowypowers: if this is static, it affects the entire class which doesnt make sense?

packages/neon-core/src/tx/transaction/PublishTransaction.ts
dauTT: Thanks for the hint. I will leverage whatever we already have.

packages/neon-core/src/tx/transaction/PublishTransaction.ts
dauTT: I can remove the `static` word but if I do, I will get a lot of warnings like this one:

> Declaration of public static method not allowed after declaration of public instance field. Instead, this should come at the beginning of the class/interface. (member-ordering)

Actually initally I just wanted to use the `version` from the `BaseTransaction` via the `this` keyword. But this is not possible because the `version` in the `BaseTransaction` is declared as `public` and
I can't use it in the `static` method `deserializeExclusive` of the subclass `PublishTransaction`.

snowypowers: The way that `deserializeExclusive` is used assumes 2 things:
1. The StringStream passed in has the pointer pointing at byte that represents the start of the exclusive data
2. The partial transaction passed in has the previous data deserialized and filled in.

We should be checking the `version` field of the transaction passed in for the version.

hal0x2328:
300 points dauTT

lllwvlvwlll:
https://neoscan.io/transaction/589328c622e101065f6717d61a332f93aac71cc68b665e194ef8c874a7e26669

PR #438, opened 2019-06-26
Update API Reference Links In Index File

uvmetal:
While reviewing this for a video demonstration I noticed the wrong links were there for the API references. These have been updated.


snowypowers:
please correct your master. there is a lot of issues with your master that is messing up your contribution. I would prefer if you actually opened a feature branch so the next time i help people with edits I dont also mess up their master.

PR #439, opened 2019-06-26
Update API Reference Links In Index File 

uvmetal:
Updated per snowy's request. This PR fixes the API doc reference links in the index for the browser demo jsfiddle code.


package.json
```diff
@@ -41,5 +41,8 @@
   "private": true,
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "@cityofzion/neon-js": "^4.6.0"
```
snowypowers: we cannot depend on ourselves

examples/browser/README.md
```diff
@@ -186,7 +186,7 @@ getrawmempool()
 
 Gets a list of unconfirmed transactions in memory.
 
-JSFiddle: [getrawmempool](https://jsfiddle.net/gh/get/jquery/3.0/cityofzion/neon-js/tree/master/examples/browser/getrawmempool/demo)
+JSFiddle: [getrawmempool](https://jsfiddle.net/gh/get/jquery/3.0/phetter/neon-js/tree/master/examples/browser/getrawmempool/demo)
```
snowypowers: why do you want to revert it back to relying on `phetter` namespace as compared to `cityofzion` namespace?

uvmetal: Thanks for catching this. I will resolve. That was from the old test code.

uvmetal:
Closed to fix in another branch.

PR #440, opened 2019-06-27
NEP9 Execute as transaction method

MorganDream:
1. add 2.10.2 rpc methods
2. nep9 send transaction native and nep5 assets


MorganDream:
@snowypowers Would you like to review this?

snowypowers:
You are adding 2 separate features in the PR. Lets split it up.

- For v2.10.2 methods you also be adding methods to the RPCClient. Do add the corresponding tests too.

- For NEP9, my personal preference is to actually keep it light and not provide any sort of execution at all. The module was solely meant to parse the uri string into something that was more easily processed.

- I do agree that that is quite a bit of code to get the NEP9Intent properly executed. My suggestion would be to have the method instead take in a Balance object and the intent and output a Transaction object instead. I would prefer not to mangle with RPC calls within this module if I can help it.

MorganDream:
@snowypowers Thanks for your reviewing!
> You are adding 2 separate features in the PR. Lets split it up.
> 
> * For v2.10.2 methods you also be adding methods to the RPCClient. Do add the corresponding tests too.
>> OK, will do it.
> * For NEP9, my personal preference is to actually keep it light and not provide any sort of execution at all. The module was solely meant to parse the uri string into something that was more easily processed.
>> I think SDK should provide customized high-level module that would be easy for developers to invoke. That's what a SDK should accomplish.
> * I do agree that that is quite a bit of code to get the NEP9Intent properly executed. My suggestion would be to have the method instead take in a Balance object and the intent and output a Transaction object instead. I would prefer not to mangle with RPC calls within this module if I can help it.
>> Actually my first thought is to output a Transaction object from a NEP9 intent. To construct a transaction, you need to get Balance object through RPC(or neoscan/neodb), which is specific according to mainNet/testNet, and the transaction also needs a RPC to be executed. 
This process would require user to specify the RPC url twice, which is not proper design, and easy to make mistakes.
So I decide to let user input once the RPC url to execute the NEP9 transaction. I have also considered to use apiProvider, but I see in nep5 module, it uses RPC module. And using apiProvider means importing another package. So I finally decided to use RPC, to be consistent with NEP5.

snowypowers:
- You have not added the RPCClient methods for the new v2.10.2 methods. Adding those methods will require you to statically type the return values from the endpoints and that would be more work than just constructing the Query objects. That will increase the PR size and thus please split this PR into 2 as I have suggested.

- I actually havnt given the new methods too much thought because the issue is that they exist on a plugin and thus are not always available on every neo node running even if the version is v2.10.2. This complicates the implementation a little. Im not sure how should we surface this.

- NEP5 and NEP9 are very different NEPs. NEP5 describes a static interface for a contract on the blockchain and thus, providing readonly methods is very simple. However, the method you are proposing is a write method onto the blockchain. This comes with many complications (eg. Ledger signing). In this case, I think deviating from the preset is okay. I think it is fine for the user to pass in the Balance object. It is up to the user to pull the Balance object from the appropriate source (neoscan or neo node).

tl;dr: 

- Please split the PR. Im insistent on this.
- Please add the RPCClient methods for the 3 methods. Im insistent on this.
- The design for NEP9 is up for discussion and it would be better if we had issues opened for this instead of committing code without a heads up.

MorganDream:
@snowypowers I agree with you.
Have splitted the PR, this PR only contains NEP 9 modification.
RPC methods are moved to https://github.com/CityOfZion/neon-js/pull/443.

PR #442, opened 2019-06-28
docs(core): Fix cdn script tag in examples

uvmetal:
1. Fix broken references to docs.neo.org 
2. Update jsfiddles to neon-js 4.6 cdn


examples/browser/getassetstate/demo/demo.html
```diff
@@ -1,4 +1,4 @@
-<script src="https://cdn.jsdelivr.net/npm/@cityofzion/neon-js@4.5.2/dist/browser.min.js"></script>
```
snowypowers: how about using a non-versioned url like `https://cdn.jsdelivr.net/npm/@cityofzion/neon-js` to always get the latest

uvmetal: Will do, thanks for the support.

PR #443, opened 2019-07-03
Rpc methods of 2.10.2

MorganDream:
@snowypowers integration test for RPCClients fails, I guess it's because coz test net node has not installed plugin RpcSystemAssetTracker. 
What do you suggest?

### DIFF
--- a/packages/neon-core/src/rpc/RPCClient.ts
+++ b/packages/neon-core/src/rpc/RPCClient.ts
@@ -324,7 +324,7 @@ export class RPCClient {
    */
   public async getUnspents(addr: string): Promise<any> {
     const response = await this.execute(Query.getUnspents(addr));
-    return response.result.balance;
+    return response.result;
   }
 
   /**
@@ -340,7 +340,7 @@ export class RPCClient {
    */
   public async getClaimable(addr: string): Promise<any> {
     const response = await this.execute(Query.getClaimable(addr));
-    return response.result.claimable;
+    return response.result;
   }
 }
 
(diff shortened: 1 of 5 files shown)


snowypowers:
Since this is a readonly endpoint we can point the tests at mainNet. We should notify COZ to upgrade their nodes too.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -318,6 +318,30 @@ export class RPCClient {
     const response = await this.execute(Query.validateAddress(addr));
     return response.result.isvalid;
   }
+
+  /**
+   * Get the unspent utxo for an address
+   */
+  public async getUnspents(addr: string): Promise<any> {
```
snowypowers: Can we strongly type these return values? I know the old RPC methods are not typed but lets start typing the new ones. It would be easier to work with them. For balance, you will need to pull in a parser (pretty close to the neoscan implementation).

MorganDream: @snowypowers Good idea.
How about typing all RPC methods in neo3 version of neon-js?
For neo2, let's just keep it working like this.
What do you say?

snowypowers: lets just implement it correctly here.

MorganDream: @snowypowers OK. Done.

MorganDream:
This is closed for targetting `master` branch, New PR: https://github.com/CityOfZion/neon-js/pull/446

PR #449, opened 2019-07-16
Neo3 development - neon-core - tx, wallet

MorganDream:
# NEO3 wallet & tx modules
## Native Contracts
In NEO3, UTXO model is abandoned, all assets become contracts including NEO & GAS. NEO, GAS and a new contract called `POLICY` are called `Native Contracts`, which are native to neo system.
- NEO, nep5 standard, also has functions related to `claiming gas` and `validators`
- GAS, nep5 standard, also has function `GetSysFeeAmount`
- POLICY, has functions about policy and validators, mainly used by consensus nodes.

### Transaction Type
All removed and only one type left, or no type. All transactions are like `InvocationTransaction` in NEO2.
### Version
Must be 0.
### Attributes
Only 2 usage are left: `Cosigner` & `Url`.
Actually `Cosigner` will be removed, but we have not changed this is this PR.
### UTXO, inputs, outputs
All gone.
### New Props
- <b>nonce</b>: 4-bytes unsigned integer to avoid hash collision, which is like that in Bitcoin.
- <b>sender</b>: The payer of systemFee and networkFee
- <b>systemFee</b>: The gas cost for running script in VM, distributed to NEO holders, 
- <b>networkFee</b>: The gas cost for transaction verification and size, distributed to CN
- <b>validUntilBlock</b>: Current transaction will be invalid after block of height validUntilBlock

### Balance & Claim
Removed
### Account
Your address is changing!
`VerifcationScript = OpCode.PUSHBYTES33 + publicKey + OpCode.SYSCALL + InteropServiceCode.NEO_CRYPTO_CHECKSIG`
So the addresses derived from same private key in NEO2 and NEO3 are different.
### Multi-Signature
Use `InteropServiceCode.NEO_CRYPTO_CHECKMULTISIG` instead of `Opcode.CHECKMULTISIG`


MorganDream:
@snowypowers Now the low level API is finished, so is the related ut.
Could you review and give suggestions?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
+   * networkFee = verificationCost + txSize * POLICY_FEE_PERBYTE
+   * @param autoAjust when true, auto justify tx networkfee to the minimum value
+   * @returns minimum networkFee that this tx needs
+   */
+  public calculateNetworkFee(autoAdjust: boolean = true): Fixed8 {
+    const signers = this.getScriptHashesForVerifying();
+    // compute the size of transaction.
+    const txHeaderSize =
+      /* version */ 1 +
+      /* nonce */ 4 /* sender */ +
```
snowypowers: is the nonce always a fixed size?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
+    }
+  }
+
+  /**
+   * calculate networkFee
+   * networkFee = verificationCost + txSize * POLICY_FEE_PERBYTE
+   * @param autoAjust when true, auto justify tx networkfee to the minimum value
+   * @returns minimum networkFee that this tx needs
+   */
+  public calculateNetworkFee(autoAdjust: boolean = true): Fixed8 {
```
snowypowers: refactor this method. Its unreadable.

packages/neon-core/src/wallet/multisig.ts
```diff
@@ -51,11 +51,11 @@ export function getSigningThresholdFromVerificationScript(
   verificationScript: string
 ): number {
   const checkSigOpCode = verificationScript.slice(
```
snowypowers: rename var to something like InteropCode instead.

packages/neon-core/src/wallet/core.ts
```diff
@@ -87,7 +95,12 @@ export function getPublicKeyFromPrivateKey(
 export const getVerificationScriptFromPublicKey = (
   publicKey: string
 ): string => {
-  return "21" + publicKey + "ac";
+  return (
+    num2hexstring(OpCode.PUSHBYTES33) +
+    publicKey +
+    num2hexstring(OpCode.SYSCALL) +
+    InteropService.NEO_CRYPTO_CHECKSIG
```
snowypowers: This really reveals the differences in implementation. Both are supposed to be similar in nature but here the usage is really awkward with one requiring a num2hexstring while the other is native.

packages/neon-core/src/tx/transaction/main.ts
```diff
@@ -23,51 +25,87 @@ export function serializeArrayOf(prop: any[]): string {
...
 }
 
-export function deserializeInputs(
+export function deserializeScript(
+  ss: StringStream,
+  tx: Partial<TransactionLike> = {}
+): Partial<TransactionLike> {
+  const script = ss.readVarBytes();
+  if (script.length === 0) {
+    throw new Error("Script should not be vacant.");
```
snowypowers: Unless this is an illegal state, we should not throw anything. Let it deserialize.

packages/neon-core/src/tx/transaction/main.ts
```diff
@@ -23,51 +25,87 @@ export function serializeArrayOf(prop: any[]): string {
...
+    ss
+  ).map(i => i.export());
+  const cosigners = getCosignersFromAttributes(attributes);
+  if (
+    !cosigners.every(
+      cosigner =>
+        cosigners.indexOf(cosigner) === cosigners.lastIndexOf(cosigner)
+    )
+  ) {
+    throw new Error("Cosigner should not duplicate.");
```
snowypowers: Same as above, if this is an deserialization issue, lets not throw here. We can do validation at another layer, perhaps when we are transforming the TransactionLike model into the Transaction object. Separating our concerns allows for reusability too.

packages/neon-core/src/tx/transaction/main.ts
```diff
@@ -79,3 +117,31 @@ export function deserializeWitnesses(
...
+    ensureHex(sender);
+    return sender;
+  } else if (sender.length === 34) {
+    return getScriptHashFromAddress(sender);
+  } else {
+    throw new Error(`Sender format error: ${sender}`);
+  }
+}
+
+export function getVarSize(value: number): number {
```
snowypowers: This looks suspiciously like a util method. Have we changed how we read var bytes?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
+   * A random number to avoid hash collision
+   */
+  public nonce: number;
+  /**
+   * transation invoker in script hash
+   */
+  public sender: string;
+
+  /**
+   * Distributed to NEO holders
```
snowypowers: The developer won't care about the distribution. What is this fee for and how is it calculated.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
+   */
+  public systemFee: Fixed8;
+
+  /**
+   * Distributed to consensus nodes
```
snowypowers: Same here. I dont care about distribution. I care about how do I arrive at the correct amount.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
+   */
+  public networkFee: Fixed8;
+
+  /**
+   * System Fee calculated while adding intents
+   */
+  private _pre_systemFee: Fixed8;
+
+  /**
+   * Current transaction will not be valid until block of height validUntilBlock
```
snowypowers: This is confusing. I thought this was an expiry timestamp, aka validBeforeBlockIsReached. Your description suggests invalidBeforeBlockIsReached.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
-  public get [Symbol.toStringTag](): string {
-    return "Transaction";
+  /**
+   * Adds some script intents to the Transaction
+   * System Fee will be increased automatically
+   * However system Fee calculated in this method is insufficient if calling non-native contracts.
+   * If the transaction is invoking contracts other than native contracts, invokeScript rpc request can test the systemFee.
+   * @param scriptIntents sciprt Intents to add to the transaction
+   */
+  public addIntent(...scriptIntents: ScriptIntent[]): this {
```
snowypowers: The method should be addIntents

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,386 @@
...
   }
 
-  public get fees(): number {
-    return 0;
+  /**
+   * This function regulate system fee with 2 processes:
+   * 1. considering the free system fee threshold
+   * 2. ceil systemFee with factor
+   */
+  public regulateSystemFee(): void {
```
snowypowers: please refactor + explain the logic more.

packages/neon-core/src/rpc/Query.ts
```diff
@@ -157,11 +152,15 @@ export class Query {
   }
 
   /**
-   * This Query returns the transaction hashes of the transactions waiting to be processed at the node.
+   * This Query returns the transaction hashes of the transactions confirmed or unconfirmed.
+   * @param shouldGetUnverified Optional. Default is 0.
+   * shouldGetUnverified = 0, get confirmed transaction hashes
```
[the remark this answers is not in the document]
MorganDream: Yes when `shouldGetUnverified=1`.
https://github.com/Tommo-L/NEO3-Development-Guide/blob/master/en/RPC/api/getrawmempool.md

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Yes. I think so.
```c#
public const int HeaderSize =
            sizeof(byte) +  //Version
            sizeof(uint) +  //Nonce
```
Above is part of the code to calculate tx header size.

packages/neon-core/src/wallet/core.ts
MorganDream: Yes. Maybe I should change InteropSerivce to type number instead of HEX?

packages/neon-core/src/tx/transaction/main.ts
MorganDream: https://github.com/neo-project/neo/blob/2ef5d34cd88facbdfd13fd81edd8ad25012879b3/neo/Network/P2P/Payloads/Transaction.cs#L98
The core code throws.
I think it's because a transaction cannot be vacant.

packages/neon-core/src/tx/transaction/main.ts
MorganDream: https://github.com/neo-project/neo/blob/2ef5d34cd88facbdfd13fd81edd8ad25012879b3/neo/Network/P2P/Payloads/Transaction.cs#L96

MorganDream: But as you said, we could do validation in other parts. If you still think so considering the core code, I will do it.

packages/neon-core/src/tx/transaction/main.ts
MorganDream: This is another method to calculate the compressed size according to number value. It's used in calculating networkFee according to tx size. The core code use this method name, maybe we can change it to another name to avoid confusion?

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: `systemFee` will be distributed to neo holders as claimable gas.
system fee.
It will be payed by `sender`.
And this prop is necessary in tx structure.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: We must keep these 2 props, unlike NEO2 with utxo, `systemFee, networkFee` will be written into transaction when serialized in NEO3.
Maybe we could consider combine them in high level API.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Yes. You are right, will modify desc.

packages/neon-core/src/wallet/multisig.ts
MorganDream: Do you mean the `8` in `verificationScript.length - 8`?

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: explanation updated.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: refactored

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Changed

packages/neon-core/src/tx/transaction/main.ts
snowypowers: The core code expects fully formed transactions. We are trying to construct it here so I think its fine to let it go empty.

packages/neon-core/src/tx/transaction/main.ts
snowypowers: yeah i think its fine to not do validation at the deserialization step unless the data is badly malformed.

packages/neon-core/src/tx/transaction/main.ts
snowypowers: yes. And why are we exporting this? Where else will we use it other than in Transaction?

packages/neon-core/src/wallet/core.ts
snowypowers: I think it might be better if we change the enum OpCode to use string since we dont really do arithmetic on the opcodes anyway. And the name InteropServiceCode would be more suitable since InteropService sounds like an implementation not an enum.

packages/neon-core/src/wallet/multisig.ts
snowypowers: This isnt checking an OpCode anymore right? Its an Interop? Unless we want to combine both classes of code under the same enum.

MorganDream: It is an Interop, not an opcode anymore. all interop hash length will be 8.

packages/neon-core/src/wallet/core.ts
MorganDream: Good, I'll try to implement this.

packages/neon-core/src/tx/transaction/main.ts
MorganDream: It's only used in Transaction. Since only one type of transaction lingers, most functions in main.ts is only used in Transaction. But moving all these methods to Transaction would make a huge class. Anyway these are not exported from module 'tx', which I think is good enough.

packages/neon-core/src/tx/transaction/main.ts
MorganDream: OK.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: OK. changed.

packages/neon-core/src/tx/txAttrUsage.ts
```diff
@@ -3,44 +3,8 @@
  * @enum {number}
```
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,377 @@
...
- * Transactions are what you use to interact with the blockchain.
- * A transaction is made up of components found in the component file.
- * Besides those components which are found in every transaction, there are also special data that is unique to each transaction type. These 'exclusive' data can be found in the exclusive file.
- * This class is a wrapper around the various transaction building methods found in this folder.
- */
-export class Transaction extends BaseTransaction {
+export class Transaction {
+  public version: number;
+  /**
+   * A random number to avoid hash collision
```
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-core/src/sc/NativeContractMethodPrices.ts
```diff
@@ -0,0 +1,27 @@
+import { Fixed8 } from "../u";
+
+export const NativeContractMethodPrices = new Map<string, Fixed8>([
```
snowypowers: any particular reason you chose a map instead of just a generic object?

packages/neon-core/src/sc/NativeContractMethodPrices.ts
```diff
@@ -0,0 +1,27 @@
...
+  ["getBlockedAccounts", new Fixed8(1000000e-8)],
+  ["setMaxTransactionsPerBlock", new Fixed8(3000000e-8)],
+  ["setFeePerByte", new Fixed8(3000000e-8)],
+  ["blockAccount", new Fixed8(3000000e-8)],
+  ["unblockAccount", new Fixed8(3000000e-8)],
+  ["supportedStandards", new Fixed8(0)],
+  ["name", new Fixed8(0)],
+  ["symbol", new Fixed8(0)],
+  ["decimals", new Fixed8(0)],
+  ["totalSupply", new Fixed8(1000000e-8)],
```
snowypowers: all these seem to be the NEP5 token contract. Is there going to be only 1 native contract forever? Or should we rename this to NativeTokenContractMethodPrices

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,379 @@
...
+  /**
+   * Current transaction will be invalid after block of height validUntilBlock
+   */
+  public validUntilBlock: number;
+  public attributes: TransactionAttribute[];
+  public intents: ScriptIntent[];
+  public scripts: Witness[];
+  public script: string;
+
+  public static MAX_VALIDUNTILBLOCK_INCREMENT = 2102400;
```
snowypowers: can we rename this to either MAX_TRANSACTION_LIFESPAN or MAX_TRANSACTION_VALIDITY? You lose the meaning by putting increment here. You can add a comment directing it back to the C# name if you wish.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,379 @@
...
+      intents,
+      scripts,
+      script
+    } = tx;
+    this.version = version || TX_VERSION;
+    this.nonce = nonce || parseInt(ab2hexstring(generateRandomArray(4)), 16);
+    this.sender = formatSender(sender);
+    // TODO: The default should be snapshot.height + MAX_VALIDUNTILBLOCK_INCREMENT, but it needs request to get block height, thus this is a temporary value
+    this.validUntilBlock =
+      validUntilBlock || Transaction.MAX_VALIDUNTILBLOCK_INCREMENT;
```
snowypowers: default should be a illegal value like 0 / -1 so we can quickly check its validity. Dont try to magic it in the constructor.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,379 @@
...
+      ? attributes.map(a => new TransactionAttribute(a))
+      : [];
+    this.scripts = Array.isArray(scripts)
+      ? scripts.map(a => new Witness(a))
+      : [];
+    this.systemFee = systemFee ? new Fixed8(systemFee) : new Fixed8(0);
+    this.networkFee = networkFee ? new Fixed8(networkFee) : new Fixed8(0);
+    this._pre_systemFee = new Fixed8(0);
+    this.script = script || "";
+    if (intents !== undefined) {
```
snowypowers: I dont remember intents in the old object. Is this new?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,379 @@
...
+    this.nonce = nonce || parseInt(ab2hexstring(generateRandomArray(4)), 16);
+    this.sender = formatSender(sender);
+    // TODO: The default should be snapshot.height + MAX_VALIDUNTILBLOCK_INCREMENT, but it needs request to get block height, thus this is a temporary value
+    this.validUntilBlock =
+      validUntilBlock || Transaction.MAX_VALIDUNTILBLOCK_INCREMENT;
+    this.attributes = Array.isArray(attributes)
+      ? attributes.map(a => new TransactionAttribute(a))
+      : [];
+    this.scripts = Array.isArray(scripts)
+      ? scripts.map(a => new Witness(a))
```
snowypowers: witness has sorting logic. We should sort it as we append them.

packages/neon-core/src/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
snowypowers: I dont see any updates.

packages/neon-core/src/tx/transaction/Transaction.ts
snowypowers: I dont see it.

packages/neon-core/src/tx/transaction/Transaction.ts
snowypowers: is there not a ceiling method?

packages/neon-core/src/tx/transaction/main.ts
```diff
@@ -23,51 +35,87 @@ export function serializeArrayOf(prop: any[]): string {
...
-export function deserializeType(
+export function deserializeVersion(
   ss: StringStream,
   tx: Partial<TransactionLike> = {}
 ): Partial<TransactionLike> {
   const byte = ss.read();
-  return Object.assign(tx, { type: parseInt(byte, 16) });
+  const version = parseInt(byte, 16);
+  if (version !== 0) {
+    log.warn(`Transaction version should be 0 not ${version}`);
```
snowypowers: why warning here? I believe the version determines how we deserialize this so if we are encountering an unknown version we should not be able to deserialize this correctly and throw an error.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,379 @@
...
+   * networkFee = verificationCost + txSize * POLICY_FEE_PERBYTE
+   * @param autoAjust when true, auto justify tx networkfee to the minimum value
+   * @returns minimum networkFee that this tx needs
+   */
+  public calculateNetworkFee(autoAdjust: boolean = true): Fixed8 {
+    let networkFee = new Fixed8(0);
+
+    const signers = this.getScriptHashesForVerifying();
+
+    let size =
```
snowypowers: why cant we use the serialize function to check size?

packages/neon-core/src/wallet/multisig.ts
snowypowers: so the name should be checkInteropCode, not checkSigOpCode

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Sorry, it's changed for sure now. Please check again.

packages/neon-core/src/wallet/multisig.ts
MorganDream: OK. Changed.

packages/neon-core/src/sc/NativeContractMethodPrices.ts
MorganDream: The keys are native contract methods, which are strings. Keys of enum and generic object are usually fixed prop name, and better in upper case. That's why I chose Map.

packages/neon-core/src/sc/NativeContractMethodPrices.ts
MorganDream: > Is there going to be only 1 native contract forever? 
I don't think I get what you mean, there are 3 native contracts now: NEO, GAS, POLICY, there would possibly be more in future. These contracts have some same-name methods, and their prices are same. That's why I use this to map price to method but not to contract.
>  should we rename this to NativeTokenContractMethodPrices
I think not. Because POLICY contracts are not tokens.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: This is new. And it's not from the core code.
I add it because user can use it to let sdk build script. And specifying it as prop so that user can track what will the transaction do as script is unreadable.

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Actually using serialize function to check size is my first thought.
But after checking the core code, it didn't use serialize function. And it uses the compressed size (that's why there is a function `getCompressedSize`) of transaction(the transaction will be compressed in the p2p communication).
I'm not sure if the serialize function will return the same as this.

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -1,31 +1,265 @@
...
-    expect(result.scripts.length).toBe(1);
+  test("addWitness", () => {
+    tx1.addWitness(
+      new Witness({
+        invocationScript: "ab",
+        verificationScript:
+          "210317595a739cfe90ea90b6392814bcdebcd4c920cb149d0ac2d88676f1b0894fba68747476aa"
+      })
+    );
+    expect(tx1.scripts[0].invocationScript).toBe("ab");
```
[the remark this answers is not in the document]
MorganDream: Add the verificationScript check.
I wonder what's its referential, do you mean the publicKey in the verification script?

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -1,31 +1,265 @@
...
+    systemFee: 12,
+    networkFee: 13,
+    validUntilBlock: 1000,
+    attributes: [],
+    scripts: [],
+    script: "00"
+  });
+
+  test("addAttribute", () => {
+    tx1.addAttribute(32, "9b58c48f384a4cf14d98c97fc09a9ba9c42d0e26");
```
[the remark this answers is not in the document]
MorganDream: OK.

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -1,31 +1,265 @@
...
+    nonce: 123,
+    sender: "39e9c91012be63a58504e52b7318c1274554ae3d",
+    systemFee: 12,
+    networkFee: 13,
+    validUntilBlock: 1000,
+    attributes: [],
+    scripts: [],
+    intents: [{ scriptHash: "NEO", operation: "transfer", args: [10] }]
+  });
+  test("calculateNetworkFee", () => {
```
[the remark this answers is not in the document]
MorganDream: Added

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -1,31 +1,265 @@
...
+    tx1.calculateNetworkFee(true);
+    expect(tx1.networkFee.toNumber()).toBe(0.0115024);
+  });
+
+  test("useCalculatedSystemFee", () => {
+    tx1.useCalculatedSystemFee();
+    expect(tx1.systemFee.toNumber()).toBe(0.0800718);
+  });
+
+  test("", () => {
```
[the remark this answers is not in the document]
MorganDream: Test name added.
User can get minimum system fee calculated using some method while constructing tx script, regulateSystemFee will do:
1. minus the free fee threshold(it's 10 gas in NEO2)
2. Ceil the system fee to be multiples of systemFee factor, which is now 1 gas.

packages/neon-core/src/consts.ts
```diff
@@ -2,14 +2,18 @@ export const ADDR_VERSION = "17";
 
 export const ASSETS: { [key: string]: string } = {
```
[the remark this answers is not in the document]
MorganDream: Why should we change the names?

packages/neon-core/src/consts.ts
```diff
@@ -86,14 +83,12 @@ export const NEP_FLAG = "e0";
...
-  REGISTER: 0
-};
+// transaction related
+export const TX_VERSION = 0;
+
+export const SYSTEM_FEE_FACTOR = 1;
+
+export const SYSTEM_FEE_FREE = 0;
+
+// this can be updated by query policy native contract method `getFeePerByte`
```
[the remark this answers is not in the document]
MorganDream: Should not describe it like this, my bad.
I mean user can also get this value by invoke policy contract.

This value can be updated only when a update to neo-cli happens, so it's ok to set it as const.

packages/neon-core/src/sc/InteropServiceCode.ts
```diff
@@ -0,0 +1,72 @@
+export enum InteropServiceCode {
```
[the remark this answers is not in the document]
MorganDream: No, no sorting order.
This is exactly the order from core code. Shall we change it to alphabetical?

packages/neon-core/src/sc/InteropServicePrices.ts
```diff
@@ -0,0 +1,137 @@
...
+
+function getStoragePrice(size: number): number {
+  return size * gasPerByte;
+}
+
+function getCheckMultiSigPrice(size: number): number {
+  return size * fixedPrices[InteropServiceCode.NEO_CRYPTO_CHECKSIG];
+}
+
+export function getInteropSericePrice(
```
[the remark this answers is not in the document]
MorganDream: Changed

packages/neon-core/src/sc/InteropServicePrices.ts
```diff
@@ -0,0 +1,137 @@
+import InteropServiceCode from "./InteropServiceCode";
+import NativeContractMethodPrices from "./NativeContractMethodPrices";
+import { Fixed8 } from "../u";
+
+const gasPerByte = 100000e-8;
```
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-core/src/sc/InteropServicePrices.ts
```diff
@@ -0,0 +1,137 @@
...
+      ).toNumber();
+    case InteropServiceCode.NEO_NATIVE_POLICY:
+      if (method === undefined) {
+        throw new Error(
+          `method param is necessary for InteropServiceCode ${service}`
+        );
+      }
+      return (
+        NativeContractMethodPrices.get(method) || new Fixed8(0)
+      ).toNumber();
```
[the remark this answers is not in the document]
MorganDream: Changed

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: Good idea. And Changed

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: ok, changed to 0

packages/neon-core/src/tx/transaction/Transaction.ts
MorganDream: OK. Changed

packages/neon-core/src/tx/transaction/main.ts
MorganDream: Changed

packages/neon-core/__tests__/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
snowypowers: when you do a addWitness should we be creating a new object or just pushing the object into the array. 
```
var witness = new Witness();
tx.addWitness(witness);
var valueEquality = witness.equal(tx.scripts[0]); //definitely true
var referentialEquality = witness === tx.scripts[0]; // true or false?
```

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -1,31 +1,290 @@
...
+      networkFee: 0,
+      validUntilBlock: 1000,
+      attributes: [],
+      scripts: [],
+      intents: [{ scriptHash: "NEO", operation: "transfer", args: [10] }]
+    });
+  }
+
+  describe("calculateNetworkFee", () => {
+    test("true as param to assign networkFee", () => {
```
snowypowers: Why do we want to have this boolean? If we have cases to calculate the fees without affecting the transaction, then the method should be extracted out as a static method and the transaction class calling that method.

packages/neon-core/src/consts.ts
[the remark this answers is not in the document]
snowypowers: are they still called assets?

packages/neon-core/src/sc/InteropServiceCode.ts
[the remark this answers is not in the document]
snowypowers: alphabetical is good.

packages/neon-core/src/sc/NativeContractMethodPrices.ts
snowypowers: but this class suggests that i can call totalSupply, balanceOf and all the method on Policy contract since it is considered a native contract. Or is Policy contract not native?

snowypowers: And why are we adding sc within a tx, wallet PR

packages/neon-core/src/consts.ts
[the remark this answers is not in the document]
MorganDream: Oh, Policy and the coming Vote contracts are not assets any more. Will change the name to NATIVE_CONTRACTS?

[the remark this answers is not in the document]
MorganDream: Oh, Policy and the coming Vote contracts are not assets any more. Will change the name to NATIVE_CONTRACTS?

packages/neon-core/__tests__/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
MorganDream: Will create a new object while `addWitness`

packages/neon-core/src/sc/NativeContractMethodPrices.ts
MorganDream: Policy is native.
I see what you mean. How about not exporting this class from `sc` module?
> And why are we adding sc within a tx, wallet PR
Because in tx has methods to calculate systemFee and it needs the support from `sc`

MorganDream:
> why is there so much change to the sc folder in this PR which is supposed to be tx and wallet?

@snowypowers Mostly it's because the calculation of systemFee while `addIntents` to transaction.
System Fee is the price of opcodes and interopservice codes, it can be calculated during the construction of script, which can be done in `ScriptBuilder` class.  Please reconsider about this.

packages/neon-core/__tests__/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
MorganDream: Changed

packages/neon-core/__tests__/tx/transaction/Transaction.ts
MorganDream: Removed this boolean param

snowypowers:
lets not add any new features while we are trying to migrate versions. There is too much uncertainty at the moment.

MorganDream:
@snowypowers OK. I have prepared a PR about sc and rpc which is small: https://github.com/CityOfZion/neon-js/pull/454. Let get it merged first.

packages/neon-core/src/consts.ts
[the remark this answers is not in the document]
MorganDream: I keep the `ASSET_ID` name to avoid too many changes

MorganDream:
@snowypowers I have resolved conflicts and now this PR is only about wallet and tx, still lots of deletions from high level api to avoid build error

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,395 @@
...
+    } else {
+      this.systemFee = new Fixed8(0);
+    }
+  }
+
+  /**
+   * calculate networkFee and assign it to tx networkFee prop
+   * networkFee = verificationCost + txSize * POLICY_FEE_PERBYTE
+   */
+  public calculateNetworkFee(): void {
```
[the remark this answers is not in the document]
MorganDream: What do you mean by no new features here? Shall I describe this in the PR description?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,395 @@
...
+   */
+  public systemFee: Fixed8;
+
+  /**
+   * Distributed to consensus nodes
+   * networkFee is calculated according to transaction size and verificationScript cost in witnesses.
+   * ```ts
+   * const networkFee = FeePerByte * txSize + SUM(verificationScriptCost)
+   * ```
+   * The calculation of `verificationScriptCost` is same as systemFee
```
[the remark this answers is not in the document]
MorganDream: Description updated. 
Actually it cannot always be calculated offline. It cannot when it's contract verification script. I have left a `TODO` in the function and wait for the core code to update.

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,395 @@
...
+  public nonce: number;
+  /**
+   * transation invoker in script hash
+   */
+  public sender: string;
+
+  /**
+   * Distributed to NEO holders
+   * systemFee is calculated by summarizing prices of all the opcodes and interopServices used while executing transaction script in vm.
+   * ```ts
```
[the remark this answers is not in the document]
MorganDream: ok

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,395 @@
...
+   * Current transaction will be invalid after block of height validUntilBlock
+   */
+  public validUntilBlock: number;
+  public attributes: TransactionAttribute[];
+  public intents: ScriptIntent[];
+  public scripts: Witness[];
+  public script: string;
+
+  /**
+   * `MAX_VALIDUNTILBLOCK_INCREMENT` in neo core
```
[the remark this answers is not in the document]
MorganDream: ok

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -1,104 +1,395 @@
...
-  public get [Symbol.toStringTag](): string {
-    return "Transaction";
+  /**
+   * Adds some script intents to the Transaction
+   * System Fee will be increased automatically (currently deprecated)
+   * However system Fee calculated in this method is insufficient if calling non-native contracts.
+   * If the transaction is invoking contracts other than native contracts, invokeScript rpc request can test the systemFee.
+   * @param scriptIntents sciprt Intents to add to the transaction
+   */
+  public addIntents(...scriptIntents: ScriptIntent[]): this {
```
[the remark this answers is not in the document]
MorganDream: I think this design is rational. As in NEO3, only invocationTransaction is left and transaction is mainly about `script`, which needs tight relationship with `sc` module.

packages/neon-core/src/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-core/src/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
snowypowers: Lets keep the data class clean first and construct the individual pieces properly before deciding on how to put things together. Not having this method also means I can tree shake and reduce package size if I only need to parse the transaction.

packages/neon-core/src/tx/transaction/main.ts
```diff
@@ -79,3 +124,67 @@ export function deserializeWitnesses(
...
+  if (value < 0xfd) {
+    return 1;
+  } else if (value < 0xffff) {
+    return 3;
+  } else {
+    return 5;
+  }
+}
+
+export function getNetworkFeeForSig(): number {
```
snowypowers: can we move all the fees related stuff out? Fees can be a different class at this point.

packages/neon-core/src/tx/transaction/Transaction.ts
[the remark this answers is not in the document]
MorganDream: Removed

packages/neon-core/src/tx/transaction/main.ts
MorganDream: OK

packages/neon-core/src/tx/transaction/main.ts
snowypowers: This is still throwing an error.

packages/neon-core/src/tx/transaction/main.ts
snowypowers: same, remove the check.

packages/neon-core/src/tx/transaction/main.ts
MorganDream: Sorry this must be a mistake I made when resolving conflicts. Changed

packages/neon-core/src/tx/transaction/main.ts
MorganDream: Removed

hal0x2328:
1000 points MorganDream

lllwvlvwlll:

PR #451, opened 2019-07-18
Fix Fixed8 overflow in nep5 transfer method

### LINKED ISSUE #450: Fixed8 overflow in nep5 transfer method (neon-js v4.6.0)
If we run this test using **neon-js v4.6.0**: 
```typescript

import Fixed8 from "../src/u/Fixed8";
import { transfer } from "../../neon-nep5/src/abi";

test("test3", () => {
  const account = new Account(
    "9ab7e154840daca3a2efadaf0df93cd3a5b51768c632f5433f86909d9b994a69"
  );

  const amtToSendTest = 922.33720369;
  const numOfDecimalsTest = 8;

  const generator = transfer(
    "9488220e8654d798f9b9cb9e74bd611ecc83fd0f",
    account.address,
    account.address,
    new Fixed8(amtToSendTest).div(Math.pow(10, 8 - numOfDecimalsTest))
  );
  const builder = generator();
});
```
we get the following errors: 
```typescript
Error: expected input to be less than 92233720368.54775807. Got input = 92233720369
```
The error happens in the `adjustedAmt` assignment, due to the fact that the `Fixed8` multiplication of `amt = 922.33720369` by 10^8 is greater than the maximum `Fixed8` value allowed which is 92233720368.54775807.

https://github.com/CityOfZion/neon-js/blob/master/packages/neon-nep5/src/abi.ts#L86

image

This error was not present in **neon-js v4.5.2** because we were more relax there and we didn't have any boundary for the `Fixed8` value. The expected `Scriptbuilder` is supposed to be:

```
ScriptBuilder {
      str: '0531e28e79151435b20010db73bf86371075ddfba4e6596f1ff35d1435b20010db73bf86371075ddfba4e6596f1ff35d53c1087472616e73666572670ffd83cc1e61bd749ecbb9f998d754860e228894',
      pter: 0 }

(31e28e7915 = reverse hex of 92233720369)
```

I believe in this case `adjustedAmt` is actually the integer representation of `new u.Fixed8(amt)`, in other words

```typescript
adjustedAmt = new BN(new u.Fixed8(amt).toHex(), 16)
```
This will fix the `nep5 transfer method`. But perhaps we should create a method in the `Fixed8` class which provide us its integer representation. Do you agree @snowypowers ?

dauTT:
Closes #450

### DIFF
--- a/packages/neon-core/__tests__/u/Fixed8.ts
+++ b/packages/neon-core/__tests__/u/Fixed8.ts
@@ -31,7 +31,7 @@ describe.each([
     expect(result.toReverseHex()).toBe(hex);
   });
 
-  test("toRawNumber2", () => {
+  test("toRawNumber", () => {
     const result = new Fixed8(num);
     expect(result.toRawNumber()).toEqual(new BN(raw_num));
   });
--- a/packages/neon-core/__tests__/u/Fixed8.ts
+++ b/packages/neon-core/__tests__/u/Fixed8.ts
@@ -1,14 +1,16 @@
 import Fixed8 from "../../src/u/Fixed8";
+import BN from "bignumber.js";
 import { reverseHex } from "../../src/u/misc";
 
 describe.each([
-  ["0000000000000000", 0],
-  ["00e1f50500000000", 1],
-  ["0100000000000000", 0.00000001],
-  ["0080c6a47e8d0300", 10000000],
-  ["5004fb711f010000", 12345.6789],
-  ["ffffffffffffffff", -0.00000001]
-])("from hexstring (%s)", (hex, num) => {
+  ["0000000000000000", 0, 0],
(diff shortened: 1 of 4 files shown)


packages/neon-core/__tests__/u/Fixed8.ts
```diff
@@ -28,6 +30,11 @@ describe.each([
     const result = new Fixed8(num);
     expect(result.toReverseHex()).toBe(hex);
   });
+
+  test("toRawNumber2", () => {
```
snowypowers: remove the 2

hal0x2328:
200 points dauTT

lllwvlvwlll:
https://neoscan.io/transaction/d82da2040b11e60522f6d409ad8f96d6842f527feb13941ce2fa787997ca06f7

Issue #460, opened 2019-08-20
[Bug] incompatible sync/async encryption interfaces between Wallet and Account class.


tsmcalister:
It seems like the the `Wallet.encrypt` and therefore `Wallet.encryptAll` functions should return `Promise<boolean>` and `Promise<boolean[]>` respectively.

The `Wallet.encryptAll` function calls the `Wallet.encrypt` function which in turn calls the async `Account.encrypt` function. See https://github.com/CityOfZion/neon-js/blob/master/packages/neon-core/src/wallet/Wallet.ts#L150

otherwise the following code wont work:
```ts
wallet.encryptAll("password")
wallet.export() // fails
```

Workaround is to loop over the `Wallet.accounts` object manually i.e something like this.:

```ts
const waitList = []
for (let acc of wallet.accounts){
    waitList.push(acc.encrypt("password"))
}
Promise.all(waitList).then(() =>
    wallet.export()
)
```

MorganDream:
Yes you are right. We should fix this.

MorganDream:
Shall we close this as PR is merged. @tsmcalister

PR #461, opened 2019-08-22
Fix wallet encrypt/decrypt to return promise

MorganDream:
refer to https://github.com/CityOfZion/neon-js/issues/460

### DIFF
--- a/packages/neon-core/src/wallet/Wallet.ts
+++ b/packages/neon-core/src/wallet/Wallet.ts
@@ -98,78 +98,60 @@ export class Wallet {
    * Attempts to decrypt Account at index in array.
    * @param index Index of Account in array.
    * @param keyphrase keyphrase
-   * @return Decryption success/failure
+   * @return Promise of the decrypted account.
    */
-  public decrypt(index: number, keyphrase: string): boolean {
+  public decrypt(index: number, keyphrase: string): Promise<Account> {
     if (index < 0) {
-      throw new Error("Index cannot be negative!");
+      return Promise.reject(`Index cannot be negative! index: ${index}`);
     }
     if (index >= this.accounts.length) {
-      throw new Error("Index cannot larger than Accounts array!");
+      return Promise.reject(
+        `Index cannot larger than Accounts array! index: ${index}`
+      );
     }
-    try {
-      this.accounts[index].decrypt(keyphrase, this.scrypt);
-      return true;
-    } catch (err) {
-      return false;
-    }
+    return this.accounts[index]
+      .decrypt(keyphrase, this.scrypt)
+      .catch(err => Promise.reject(`${err} index: ${index}`));
   }
 
   /**
    * Attempts to decrypt all accounts with keyphrase.
    * @param keyphrase
-   * @return Each boolean represents if that Account has been decrypted successfully.
+   * @return Promise of this(Wallet).
    */
-  public decryptAll(keyphrase: string): boolean[] {
-    const results: boolean[] = [];
-    this.accounts.map((acct, i) => {
-      results.push(this.decrypt(i, keyphrase));
-    });
-    log.info(
-      `decryptAll for Wallet ${this.name}: ${results.reduce((c, p) => {
-        return p + (c ? "1" : "0");
-      }, "")}`
-    );
-    return results;
+  public decryptAll(keyphrase: string): Promise<this> {
+    return Promise.all(
+      this.accounts.map((acct, i) => this.decrypt(i, keyphrase))
+    ).then(res => this);
   }
 
   /**
    * Attempts to encrypt Account at index in array.
    * @param index Index of Account in array.
    * @param keyphrase
-   * @return Encryption success/failure
+   * @return Promise of encrypted account
    */
-  public encrypt(index: number, keyphrase: string): boolean {
+  public encrypt(index: number, keyphrase: string): Promise<Account> {
     if (index < 0) {
-      throw new Error("Index cannot be negative!");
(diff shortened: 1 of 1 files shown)


packages/neon-core/src/wallet/Wallet.ts
```diff
@@ -98,78 +98,60 @@ export class Wallet {
    * Attempts to decrypt Account at index in array.
    * @param index Index of Account in array.
    * @param keyphrase keyphrase
-   * @return Decryption success/failure
+   * @return Promise of the decrypted account.
    */
-  public decrypt(index: number, keyphrase: string): boolean {
+  public decrypt(index: number, keyphrase: string): Promise<Account> {
```
snowypowers: Returning Promise<Account> is a breaking change. We should be fixing the method to return Promise<boolean> instead.

MorganDream: I don't get what you mean. You mean to return promise or not?

snowypowers: Promise<bool> would be the correct fix since that was the original intent.

snowypowers: ah i see. I was typing `Promise<bool>` but it was interpreted as markdown syntax

MorganDream: Changed.

packages/neon-core/src/wallet/Wallet.ts
```diff
@@ -98,78 +98,58 @@ export class Wallet {
...
+   * @return Promise of accounts decryption success/failure .
    */
-  public decryptAll(keyphrase: string): boolean[] {
-    const results: boolean[] = [];
-    this.accounts.map((acct, i) => {
-      results.push(this.decrypt(i, keyphrase));
-    });
-    log.info(
-      `decryptAll for Wallet ${this.name}: ${results.reduce((c, p) => {
-        return p + (c ? "1" : "0");
```
snowypowers: can we retain the log statement

MorganDream: That won't be necessary. Promise reject error info will reveal which index goes wrong. The old log may mislead user to regard it as success even if it fails for some index.

MorganDream:
I wonder why there isn't a unit test for this class.

snowypowers:
Yeah I think I missed out the tests for this class

MorganDream:
@snowypowers Unit test added

packages/neon-core/__tests__/wallet/Wallet.ts
```diff
@@ -0,0 +1,128 @@
...
+  });
+
+  test("default account", () => {
+    const wallet = new Wallet({
+      name: "test",
+      version: "1"
+    });
+    wallet.addAccount(new Account(PRIVATE_KEY1));
+    wallet.addAccount(new Account(PRIVATE_KEY2));
+    expect(wallet.defaultAccount).toBe(wallet.accounts[0]);
```
snowypowers: do we expect the insertion to be ordered? If yes, then the assertion should be to assert that the defaultAccount has the address of PRIVATE_KEY1.

packages/neon-core/__tests__/wallet/Wallet.ts
```diff
@@ -0,0 +1,128 @@
...
+      version: "1"
+    });
+    wallet.addAccount(new Account(PRIVATE_KEY1));
+    wallet.addAccount(new Account(PRIVATE_KEY2));
+    wallet.setDefault(1);
+    expect(wallet.defaultAccount).toBe(wallet.accounts[1]);
+  });
+});
+
+describe("encrypt & decrypt", () => {
```
snowypowers: - Add test asserting decryption failure will return rejected promise.
- Assert that after encryption, we are able to retrieve the encrypted key and it is of the correct format (if we know the encrypted key, we can just do a string compare here).
- Assert that after decryption, we are able to retrieve the correct private key.

packages/neon-core/__tests__/wallet/Wallet.ts
MorganDream: Changed

packages/neon-core/__tests__/wallet/Wallet.ts
MorganDream: Changed

packages/neon-core/__tests__/wallet/Wallet.ts
```diff
@@ -0,0 +1,166 @@
...
+    return wallet.decryptAll("passw0rd").then(res => {
+      expect(res).toStrictEqual([true, true]);
+      expect(wallet.accounts[0].privateKey).toBe(PRIVATE_KEY1);
+    });
+  });
+
+  test("failure", () => {
+    const wallet = new Wallet(WALLET_JSON);
+    return wallet.decryptAll("passw0rd_failre").catch(err => {
+      expect(err).toBeDefined();
```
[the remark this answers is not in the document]
MorganDream: @snowypowers Maybe I have a misunderstanding, but function `decryptAll` will reject the promise if the password is wrong. `catch` will catch the rejected error.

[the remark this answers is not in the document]
MorganDream: `expect.assertions(1)` added.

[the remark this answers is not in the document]
MorganDream: @snowypowers Thanks for telling me this.

hal0x2328:
300 points MorganDream
150 points snowypowers for review

lllwvlvwlll:

Issue #441, opened 2019-06-28
[NEO 3.0] neon-js should support the coming NEO 3.0


MorganDream:
NEO 3.0 road map is released. And hopefully NEO 3.0 testNet will come out in this July.
NEO plans to draw lots of communities and developers to develop tools and dApps on 3.0 testNet, in which SDK will act an important role.

Now NGD members are developing 3.0 C# SDK in project **neo**: https://github.com/neo-project/neo/pull/850. I think neon-js should be upgraded for 3.0 at the same time, because neon-js developers could request support from developers of C# SDK, as they could modify core codes properly to support SDK if necessary.

MorganDream:
## NEO3 Proposal for NEON-JS
### Current Structure of neon-js
![neon-js NEO3 version](https://user-images.githubusercontent.com/25920210/60576298-b7343380-9daf-11e9-9ce1-8d36c7526267.png)
### Targets:
To support NEO3 in neon-js.

### NEO3 features

- No UTXO, no inputs and outputs in transcation, transcation can be constructed staticly. But you need to pay attention to transaction hash collision.
- No transcation type, all transactions are invocationTransaction
- NEO and GAS become nep5 assets.
- No Claim operation, gas bonus distribution would be handled by node automatically.

### Timeline

- Release a version of neon-js that would support NEO2 and NEO3 before or short after NEO3 test net going live.
- Release a version of neon-js that is simplified to only support NEO3 before or short after NEO3 main net going live.

### Plan

- SDK can be supported just by RPC queries after neo 2.10.2, in case of delay in neoscan updating to NEO3, I’ll just implement the CLI/RPC api provider.
- Modules that need refactoring or rewriting for NEO3:
  - tx -- simplify tx structure, remove types
  - sc -- Enrich/Rewrite module sc to have method to construct invocationTransaction and method to send it through RPC methods. Or Add a module SmartContract to have such API.
  - nep5 -- refer or extends module sc for code reuse
  - new High level API(optional) -- like what neon-api does.

### Structure of neon-js for NEO3
![neon-js NEO3 version (1)](https://user-images.githubusercontent.com/25920210/60576310-bdc2ab00-9daf-11e9-9239-8bf30213cd42.png)

Any suggestions and comments would be appreciated.

snowypowers:
I think depending on how the native assets are reimplemented, we might consider merging nep5 into the core set of modules.

[...]

PR #464, opened 2019-08-29
feat(sc): Neo3 Contract Manifest

MorganDream:
refer to https://github.com/CityOfZion/neon-js/issues/441
# NEO3 Contract Manifest
Contract Manifest will record contract info, permissions, method abi, etc. It is usually in the format of JSON. It is expected to be generated while compiling neo3 contract(not implemented yet). And it will be used in contract deployment and updating, and also the NEO3 permission system.
### Why do we need manifest in neon-js?
- For contract deployment and updating(currently in this PR)
- For validation when user invoke contract method(Not in this PR)


snowypowers:
If there isnt a day to day use case, we can move contract deployment out to a separate module. Is there any examples of using the manifest to validate transactions? How do you envision the code to look like?

MorganDream:
@snowypowers 
> Is there any examples of using the manifest to validate transactions? 

Not yet. Currently I am thinking of using it to validate transaction intents. 
> How do you envision the code to look like?

This is where I want to have a discussion with you. 
Manifest looks like it can be used to dynamically generate abi for any contract, so that it could elevate user experience. It can even tell the permission control. 
Manifest info can be got from rpc request `getcontractstate`
It's used in contract deployment and updating, anyway it just needs manifest as a string.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
```diff
@@ -0,0 +1,21 @@
+export interface ContractFeaturesLike {
+  storage: boolean;
+  payable: boolean;
+}
+
+export class ContractFeatures {
```
snowypowers: ContractFeature.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
```diff
@@ -0,0 +1,21 @@
...
+  public hasStorage: boolean;
+  public payable: boolean;
+
+  public constructor(hasStorage?: boolean, payable?: boolean) {
```
snowypowers: - Overload the constructor to accept an object like all data classes.
- You can accept defaults instead of just optional args.

packages/neon-core/src/sc/manifest/ContractManifest.ts
```diff
@@ -0,0 +1,107 @@
...
+    this.abi = new ContractAbi(abi);
+    this.permissions = permissions.map(
+      permission => new ContractPermission(permission)
+    );
+    this.trusts = WildCardContainer.fromSerialized(trusts);
+    this.safeMethods = WildCardContainer.fromSerialized(safeMethods);
+    this.hash = this.abi.hash;
+  }
+
+  public static createDefault(hash: string): ContractManifest {
```
snowypowers: do we need a createDefault if we can just use the constructor with empty args to create it? Is there anything special with this default?

packages/neon-core/src/sc/manifest/ContractManifest.ts
```diff
@@ -0,0 +1,107 @@
...
+      features: this.features.export(),
+      abi: this.abi.export(),
+      permissions: this.permissions.map(permission => permission.export()),
+      trusts: this.trusts.export(),
+      safeMethods: this.safeMethods.export()
+    };
+  }
+
+  public toString(): string {
+    return JSON.stringify(this.export());
```
snowypowers: Do we have toString defined for every class? Else lets not have it.

packages/neon-core/src/sc/manifest/ContractMethodDescriptor.ts
```diff
@@ -0,0 +1,41 @@
...
+  ContractEventDescriptorLike
+} from "./ContractEventDescriptor";
+import { ContractParameterDefinition } from "./ContractParameterDefinition";
+
+export interface ContractMethodDescriptorLike
+  extends ContractEventDescriptorLike {
+  returnType: ContractParamType;
+}
+
+export class ContractMethodDescriptor extends ContractEventDescriptor {
```
snowypowers: generally I would advise against this inheritance if we are just sharing 2 fields. Is there a need to cast the method descriptor to a event descriptor?

packages/neon-core/src/sc/manifest/ContractFeatures.ts
MorganDream: Features actually. 2 features now in this class

packages/neon-core/src/sc/manifest/ContractManifest.ts
MorganDream: No, just this class. ContractManifest is the top class. It may be used when user wants to convert manifest to string and deploy contract

packages/neon-core/src/sc/manifest/ContractFeatures.ts
snowypowers: do we really need a class for this? Looks like just a data object that we can put within the main class then. Is there going to be any operations we can do with it or are they just flags?

packages/neon-core/src/sc/manifest/ContractManifest.ts
snowypowers: then let the user export it themselves no need for us to write a special method for this.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
MorganDream: Well I just want to use its `export` function when used in string <=> manifest conversion.
```json
"features": {
                "storage": true,
                "payable": true
            }
```

packages/neon-core/src/sc/manifest/ContractMethodDescriptor.ts
MorganDream: I just copied this from neo core code.
As far as I can see, they are just sharing 2 fields.

packages/neon-core/src/sc/manifest/ContractManifest.ts
MorganDream: Well user may not know exactly how to convert manifest to string. It's required to use json format string when deploying a contract. I think it's necessary to keep this method.

packages/neon-core/src/sc/manifest/ContractManifest.ts
MorganDream: Removed all the defaults, cause I don't see now how we can use them.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
MorganDream: Done

packages/neon-core/src/sc/manifest/ContractFeatures.ts
snowypowers: Lets just simplify it and include it as part of the main class then.

packages/neon-core/src/sc/manifest/ContractMethodDescriptor.ts
snowypowers: Let's split it and not use inheritance then.

packages/neon-core/src/sc/manifest/ContractManifest.ts
snowypowers: We can get there when we look at the API we will expose for deploying contracts. `toString` is one of the standard methods that has a default implementation so we dont want to attach different meanings to different classes if we can help it.

packages/neon-core/src/sc/manifest/WildCardContainer.ts
```diff
@@ -0,0 +1,35 @@
+export class WildCardContainer {
```
snowypowers: This class is not necessary a wildcard if it contains data? I dont think having a container class is the right way to do this. Just stick to strings instead.

packages/neon-core/src/sc/manifest/ContractPermissionDescriptor.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  public get isGroup(): boolean {
+    return !!this.group;
+  }
+
+  public get isWildcard(): boolean {
+    return !this.hash && !this.group;
+  }
+
+  private constructor(hash?: string, group?: string) {
```
snowypowers: Am I correct to say a permission must be either a hash or group? Cannot be both? This constructor is misleading. I see that we can always infer from the length so there is no need for 2 params. What is a group? Need more context before I can suggest an alternative.

MorganDream: > Am I correct to say a permission must be either a hash or group?

Yes.
> This constructor is misleading. 

Kind of.
> What is a group?

A group is a public key for a group of contracts. Usually for a Dapp or onchain service, more than 1 smart contract is needed, contract owner can group all Dapp related contracts for permission management.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
MorganDream: You mean to add a object-like prop in class `ContractManifest`? like below? 
```ts
features: {
  storage: boolean,
  payable: boolean
}
```

packages/neon-core/src/sc/manifest/ContractPermissionDescriptor.ts
snowypowers: I think for now, its easy to just accept a single string constructor and infer the meaning since its pretty clear that the length of the string would dictate the type of permission.

packages/neon-core/src/sc/manifest/ContractFeatures.ts
MorganDream: Done

packages/neon-core/src/sc/manifest/ContractManifest.ts
MorganDream: Removed

packages/neon-core/src/sc/manifest/ContractPermissionDescriptor.ts
MorganDream: OK

packages/neon-core/src/sc/manifest/ContractMethodDescriptor.ts
MorganDream: OK

packages/neon-core/src/sc/manifest/WildCardContainer.ts
MorganDream: Removed

packages/neon-core/src/sc/manifest/ContractPermission.ts
```diff
@@ -0,0 +1,24 @@
...
+  methods: "*" | string[];
+}
+
+export class ContractPermission {
+  public contract: ContractPermissionDescriptor;
+  public methods: "*" | string[];
+
+  public constructor(obj: Partial<ContractPermissionLike>) {
+    const { contract = "*", methods = "*" } = obj;
+    this.contract = new ContractPermissionDescriptor(contract);
```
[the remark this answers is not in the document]
MorganDream: > we might want to change the name of this field to `scope` if we d

I try to keep the consistence between the real manifest and the name of props in Manifest class. If we change it, user may be confused. (In manifest json got from rpc `getcontractstate`, it's called contract.) 

> Looks like we actually can just compress the descriptor logic into this class since its just simple validation logic.

Seems good. Except that we need the `isHash` `isGroup` interfaces, and it should belong to `contract` instead of the whole permission.
```ts
     const { contract, methods } = permission;
     if (contract.isHash) {
        if (contract.descriptor! !== manifest.hash) {
          return false;
        }
      } else if (contract.isGroup) {
        if (
          manifest.groups.every(group => group.pubKey !== contract.descriptor)
        ) {
          return false;
        }
      }
```

packages/neon-core/src/sc/manifest/ContractMethodDescriptor.ts
```diff
@@ -0,0 +1,38 @@
...
+  ContractParameterDefinition
+} from ".";
+
+export interface ContractMethodDescriptorLike {
+  name: string;
+  parameters: ContractParameterDefinitionLike[];
+  returnType: ContractParamType;
+}
+
+export class ContractMethodDescriptor {
```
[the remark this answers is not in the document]
MorganDream: I think it's better to be consistent with neo core, so that user could have less confusion.

packages/neon-core/src/sc/manifest/ContractPermissionDescriptor.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  public get isGroup(): boolean {
+    return !!this.descriptor && this.descriptor.length === 66;
+  }
+
+  public get isWildcard(): boolean {
+    return !this.descriptor || this.descriptor === "*";
+  }
+
+  public constructor(descriptor?: string) {
```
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-core/src/sc/manifest/ContractPermissionDescriptor.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (
+      !!descriptor &&
+      descriptor.length === 42 &&
+      descriptor.indexOf("0x") === 0
```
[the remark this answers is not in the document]
MorganDream: Yes, this is clearer but has redundant logic. I use a switch case logic to make it looks clearer.

packages/neon-core/src/sc/manifest/ContractPermission.ts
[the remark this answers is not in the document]
MorganDream: `contract` kept, code compressed

packages/neon-core/src/sc/manifest/ContractParameterDefinition.ts
```diff
@@ -0,0 +1,24 @@
+import { ContractParamType } from "../ContractParam";
```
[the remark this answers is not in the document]
MorganDream: Done

packages/neon-core/src/sc/manifest/ContractPermission.ts
```diff
@@ -0,0 +1,49 @@
...
+      default:
+        throw new Error(
+          `This is not a ContractPermissionDescriptor: ${contract}`
+        );
+    }
+  }
+
+  public export(): ContractPermissionLike {
+    return {
+      contract: this.isHash ? `0x${this.contract}` : this.contract,
```
snowypowers: When did we ever export with `0x` appended? Its inconsistent so lets not use it. It just means the string is a hexadecimal string. In fact if we are supporting it fully, we should be checking it for the public key too (since its also a hexadecimal string). I would remove all 0x support at the moment because we did not start with support and adding it right now is a mess.

packages/neon-core/src/sc/manifest/ContractPermission.ts
```diff
@@ -0,0 +1,49 @@
...
+  public contract: string;
+  public methods: "*" | string[];
+
+  public constructor(obj: Partial<ContractPermissionLike> = {}) {
+    const { contract = "*", methods = "*" } = obj;
+    this.contract = this._formatContract(contract);
+    this.methods = methods;
+  }
+
+  public get isHash(): boolean {
```
snowypowers: A getter will be called like a property: `permission.isHash`. Is this the intention? In JS the convention would be methods for this: `permission.isHash()`.

MorganDream: This is the intention.

packages/neon-core/src/sc/manifest/ContractPermission.ts
MorganDream: The `export` function of Manifest should be able to be used for contract deployment, I added `0x` just in case it cannot.

snowypowers: And why would the manifest require `0x` for hash but not public key? I think we should remove this as it is inconsistent at the moment.

MorganDream: OK, changed

hal0x2328:
500 points MorganDream

lllwvlvwlll:

PR #466, opened 2019-09-05
Implement buildIteratorScript

### LINKED ISSUE #437: [Enhancement] Support enumerator return from contract
NEP-11 has methods `tokens` and `tokensOfOwner` and `ownerOf` which should return an `enumerator` instead of an array or serialized data. neon-js should provide helper methods to invoke `enumerator`-returning operations more easily.

Currently, if I want to iterate over items returned in an enumerator, I need to build a script by hand to invoke the contract operation and loop with `Iterator.next`, adding each key/value to the stack before the end of the script execution so that they can be read as an array or object by the invoking dApp.

For example, on TestNet I have a contract that has an `enumerate` function to return some example items from storage. In order to build the script to iterate over the results, I wrote the following Python contract:
```
from boa.interop.Neo.App import RegisterAppCall
from boa.interop.Neo.Runtime import Log
from boa.interop.Neo.Iterator import Iterator

enumerate = RegisterAppCall('1e68de2d4442b868c544b4104c8fe2172f580591', 'operation', 'args')

def Main():
    token_iter = enumerate('enumerate',[])
    count = 0
    result = []
    while token_iter.next() and (count < 5):
        result.append(token_iter.Value)
        count += 1
    return result
```
This is then compiled and turned into in invocation script, which I can pass to neon-js as the `script` argument to `query.invokeScript`. When it finishes executing on the RPC node I should have up to 5 values from the enumerator now on the stack.

Live example: https://jsfiddle.net/hal0x2328/3ug45thd

This basic workflow should be implemented in a customizable way in neon-js, available either by providing a `postScript` parameter to be passed to `invokeFunction` and appended to the script before invocation, or as its own function, e.g. `invokeEnumeratorFunction`.

The `postScript` should be flexible and more complex than the example above, allowing for pagination of results but still easy to use in an invocation. For instance: 
```const postScript = buildIteratorScript(startIdx, maxResults)```

WyattMufson:
I added a nep11 package with a `buildIteratorScript` method.

It takes four arguments, `account: wallet.Account`,  `contractHash: string`, `page: number`, and `maxResults: number`.

It creates a post script for the nep11 `tokensOf` method by using the neo-boa compiler to compile a dynamically created script to an avm string. When https://github.com/CityOfZion/neo-boa/pull/119 is merged in, it can be updated to not require the use of tempfile and to just use the neo-boa builtin method.

This will resolve #437

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,7 +78,7 @@ jobs:
           at: .
       - run:
           name: Install neo-boa
-          command: yarn boa
+          command: yarn pip && yarn boa
       - run: mkdir reports
       - run:
           name: Integration Tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,6 +76,9 @@ jobs:
       - checkout
       - attach_workspace:
           at: .
+      - run:
+          name: Install neo-boa
+          command: yarn boa
       - run: mkdir reports
       - run:
           name: Integration Tests
--- a/docs/examples/nep-11.md
+++ b/docs/examples/nep-11.md
@@ -1,0 +1,27 @@
+---
+id: nep11
+title: NEP-11
+---
+
+## Tokens Of
+
+The `tokensOf` method returns an iterator object. In order to access the values you need to invoke a script that reads the values. `buildIteratorScript` can automatically generate the script for `tokensOf` using an `Account` and script hash.
+
+### Using buildIteratorScript
+
+```javascript
+const { wallet, nep11 } = require("@cityofzion/neon-js");
+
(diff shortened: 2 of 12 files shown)


WyattMufson:
This will require neo-boa to be installed. Is there a good way to add it?

snowypowers:
installing boa isnt a solution. This clearly has a use case in a web browser so we cannot have python scripts.

WyattMufson:
@snowypowers while this wouldn't work in a browser this would be a solution for server based projects

snowypowers:
i would prefer that we stick to pure JS since ed has shown its possible. Even server based projects might not want a pip dependency since we will have to ensure that py2/py3 is in PATH.

PR #467, opened 2019-09-06
feature(tx): Scoped witness

MorganDream:
# Scoped Witness
refer to https://github.com/neo-project/neo/pull/968 and https://github.com/CityOfZion/neon-js/issues/441
Cosigner attribute is removed and it's added in transaction as a prop, its type is a new class called `Cosigner`. Cosigner has a prop `scopes` to stand for witness scopes:  Global, CalledByEntry, CustomContracts, CustomGroups. 
Add scope to Cosigner instead of Witness for better scalability (Witness will be used in other cases like block verification, but Scoped Witness is just for transactions).
Just to mention, cosigner and sender can duplicate if user want to specify scope for sender.


packages/neon-core/src/tx/components/WitnessScope.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Used to constrain witness to specified scope
+ * The execution in vm of a script is kind of opaque, user could add scope on his signature to avoid abuse of signature
+ */
+export enum WitnessScope {
```
MorganDream: Hi @snowypowers , I want to discuss the design for this class with you, If we come to a consensus, I will start to implement the unit test.

snowypowers: We can also implement it as enum flags here.

packages/neon-core/src/tx/components/WitnessScope.ts
```diff
@@ -0,0 +1,48 @@
...
+   */
+  CustomContracts = 0x10,
+
+  /**
+   * Custom pubkey for group members, group can be found in contract manifest
+   */
+  CustomGroups = 0x20
+}
+
+export function extractScopes(scopesHex: string): WitnessScope[] {
```
snowypowers: Just return it as flags instead of array so return `WitnessScope`.

packages/neon-core/src/tx/main.ts
```diff
@@ -1,140 +1,134 @@
...
-  } else if (sender.length === 40) {
-    ensureHex(sender);
-    return sender;
-  } else if (sender.length === 34) {
-    return getScriptHashFromAddress(sender);
-  } else {
-    throw new Error(`Sender format error: ${sender}`);
-  }
-}
+import { num2VarInt, StringStream, fixed82num, ensureHex } from "../u";
```
snowypowers: Looks like a change in encoding please check.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,90 @@
...
+export class Cosigner {
+  /**
+   * script hash of cosigner
+   */
+  public account: string;
+  public scopes: WitnessScope[];
+  public allowedContracts: string[];
+  public allowedGroups: string[];
+  // This limits maximum number of allowedContracts or allowedGroups here
+  private maxSubitems: number = 16;
```
snowypowers: this is a const.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export class Cosigner {
+  /**
+   * script hash of cosigner
+   */
+  public account: string;
+  public scopes: WitnessScope[];
+  public allowedContracts: string[];
+  public allowedGroups: string[];
```
snowypowers: We can start thinking about typing these fields more strictly in the future.

packages/neon-core/src/tx/components/WitnessScope.ts
MorganDream: Please check the new design

packages/neon-core/src/tx/components/Cosigner.ts
MorganDream: Fixed

packages/neon-core/src/tx/components/Cosigner.ts
MorganDream: I had this thought, too.

packages/neon-core/src/tx/main.ts
MorganDream: This is a renamed/relocated file. I guess this diff is what it suppose to be, cause what if they are renamed but the content is exactly the same? I tried to solve it but cannot.

packages/neon-core/src/tx/components/WitnessScope.ts
snowypowers: https://basarat.gitbooks.io/typescript/docs/enums.html

This has a nice example on enum flags and having static enum methods. We shouldnt need a full class to manage flags.

MorganDream: Oh, didn't know that. Great.

MorganDream:
@snowypowers I suggest we get this merged first, as this is more important and related to the work after.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * @description This limits maximum number of allowedContracts or allowedGroups here
+   */
+  private readonly MAX_SUB_ITEMS: number = 16;
+
+  public constructor(signer: Partial<CosignerLike>) {
+    const { account, scopes, allowedContracts, allowedGroups } = signer;
+    this.account = account || "";
+    this.scopes = !scopes ? WitnessScope.Global : scopes & 0x11;
```
snowypowers: why do we want to modify the scope being passed in?

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.allowedContracts = allowedContracts || [];
+    this.allowedGroups = allowedGroups || [];
+  }
+
+  public get isGlobal() {
```
snowypowers: why have the getters when the user has clear access to the scopes. Let the user do the comparison we dont need such methods.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public get isCustomContracts() {
+    return (
+      (this.scopes & WitnessScope.CustomContracts) ===
+      WitnessScope.CustomContracts
+    );
+  }
+
+  public set isCustomContracts(yes: boolean) {
```
snowypowers: so this is kinda weird that we have a setter for the scope but not a method to set the allowedContracts. So this leads to cases where the user can:
```
var c = new Cosigner();
c.isCustomContracts= true;
```
which can be an illegal state judging by your export implementation (or a cosigner without scope?)

or the user may forget:
```
var c = new Cosigner();
c.allowedContracts.push(newContract);
```
this is an easy oversight.

Is there any valid use cases of the 2 code examples above? If not, then we can just provide a `addCustomContract` method which will add the contract to the array and amend the scope.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public static deserialize(ss: StringStream): CosignerLike {
+    const account = ss.read(20);
+    const scopes = parseInt(ss.read(), 16);
+
+    const readStringFromStream = (ss1: StringStream) => ss1.readVarBytes();
+
+    const allowedContracts =
+      (scopes & WitnessScope.CustomContracts) === WitnessScope.CustomContracts
```
snowypowers: I believe just `scopes & WitnessScope.CustomContracts` should be good enough.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @description This limits maximum number of allowedContracts or allowedGroups here
+   */
+  private readonly MAX_SUB_ITEMS: number = 16;
+
+  public constructor(signer: Partial<CosignerLike>) {
+    const { account, scopes, allowedContracts, allowedGroups } = signer;
+    this.account = account || "";
+    this.scopes = !scopes ? WitnessScope.Global : scopes & 0x11;
+    this.allowedContracts = allowedContracts || [];
```
snowypowers: We should make a copy of the array as opposed to referencing it.

packages/neon-core/src/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    return out;
+  }
+
+  public export(): CosignerLike {
+    return {
+      account: this.account,
+      scopes: this.scopes,
+      allowedContracts: this.allowedContracts,
```
snowypowers: export a copy of the array.

packages/neon-core/src/tx/main.ts
MorganDream: Fixed this.

MorganDream:
@snowypowers Unit test added and also some errors fixed, please check.

packages/neon-core/__tests__/tx/components/Cosigner.ts
```diff
@@ -0,0 +1,84 @@
...
+  test("constructor with obj & export", () => {
+    const cosigner = new Cosigner(fullCosigner);
+    expect(cosigner.export()).toStrictEqual(fullCosigner);
+  });
+});
+
+describe("add methods", () => {
+  test("addAllowedContracts", () => {
+    const cosigner = new Cosigner();
+    expect(!!(cosigner.scopes & WitnessScope.CustomContracts)).toBeFalsy();
```
snowypowers: There shouldnt be a need to use `!!` here. It is used primarily to transform unknown user inputs into bool but here the inputs are clear.

packages/neon-core/__tests__/tx/components/TransactionAttribute.ts
```diff
@@ -123,7 +124,9 @@ describe("deserialize", () => {
     "%s",
     (msg: string, expected: TransactionAttributeLike, data: string) => {
       const result = TransactionAttribute.deserialize(data);
-      expect(result.export()).toEqual(expected);
+      const exported = result.export();
```
snowypowers: This is a point of concern. If we `export`, do we expect to get a human-readable format or a transportable format? This might work now because Attributes has been stripped down to just URL. I think we should not be transforming the input unnecessarily unless the user states so. So here we will not transform the input when using the constructor but provide a static method like `TransactionAttribute.Url()` so its clear that you can use a human-readable url there.

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -154,13 +164,23 @@ describe("Add Methods", () => {
     });
   }
 
-  test("addAttribute", () => {
+  test("addCosigner", () => {
```
snowypowers: Do we still have any addAttribute test?

packages/neon-core/__tests__/tx/transaction/Transaction.ts
```diff
@@ -154,13 +164,23 @@ describe("Add Methods", () => {
...
+      account: "9b58c48f384a4cf14d98c97fc09a9ba9c42d0e26",
+      scopes: WitnessScope.Global
+    });
+    expect(tx1.cosigners[0].account).toBe(
       "9b58c48f384a4cf14d98c97fc09a9ba9c42d0e26"
     );
+    expect(tx1.cosigners[0].scopes).toBe(WitnessScope.Global);
+  });
+
+  test("addAttribute", () => {
```
MorganDream: Do we still have any addAttribute test?

>It isn't removed.

packages/neon-core/__tests__/tx/components/TransactionAttribute.ts
MorganDream: This is good. 
Just to mention, transaction attribute usage types will probably be extended in future. But the number of types would be little, I think we can do this.

packages/neon-core/src/tx/components/TransactionAttribute.ts
```diff
@@ -63,8 +74,9 @@ export class TransactionAttribute {
       throw new Error(`Data size too big!`);
```
snowypowers: This error doesnt hold anymore if you are storing it as ASCII instead of hex. The original design was always to store hex.

packages/neon-core/__tests__/tx/components/TransactionAttribute.ts
snowypowers: Lets assume we had the Remark attribute(which we had previously, leading to the current design). What do you think im sending?

```
var attr = new TransactionAttribute({usage: 1, data: "abcd"});
```

At this point it is unclear if I am attempting to input ASCII or hex. The code currently will assume hex since it looks like hex. This is horrible for the user in terms of design.

Instead, we can go for a static method:
```
var attr = TransactionAttribute.Remark("abcd");
```

We can be clear that this field is always ASCII and no more headaches.

MorganDream: Well I already agreed on this.

MorganDream: To conclude:
1. constructor will regard user input as hex, a hex checking will be included.
2. Static methods will be used to accept ASCII string.
Still this question is not answered.
> This is a point of concern. If we export, do we expect to get a human-readable format or a transportable format?
I vote for transportable format.

snowypowers: Ok then we should change the constructor to accept hex and export to return the hex value for data.

packages/neon-core/__tests__/tx/components/Cosigner.ts
MorganDream: Removed

packages/neon-core/__tests__/tx/components/TransactionAttribute.ts
MorganDream: Updated

packages/neon-core/src/tx/components/TransactionAttribute.ts
```diff
@@ -44,20 +51,43 @@ export class TransactionAttribute {
...
   }
 
   public get [Symbol.toStringTag](): string {
     return "TransactionAttribute";
   }
 
+  /**
+   * @description data in readable ASCII format
+   */
+  public get dataReadable(): string {
```
snowypowers: Please stop adding random helper methods.

MorganDream: Removed

hal0x2328:
400 points MorganDream

lllwvlvwlll:

PR #468, opened 2019-09-19
Neo3 integration test

MorganDream:
Now Neo3 testnet Preview-1 is live. So I add integration test, and fixed some bugs found when writing it:
1. sender and account in cosigner should be in big endian
2. nonce and validUntilBlock should be in little endian when serialized
3. increase robust for srciptParser

some updates:
rpcClient methods refatoring


packages/neon-core/src/rpc/Query.ts
```diff
@@ -167,9 +167,9 @@ export class Query {
   /**
    * This Query returns information about a specific transaction in either hexstring or human readable JSON.
    * @param txid hash of the specific transaction.
-   * @param verbose 0 for hexstring, 1 for JSON. Defaults to 1.
+   * @param verbose 0 for hexstring, 1 for JSON. Defaults to 0.
    */
-  public static getRawTransaction(txid: string, verbose: number = 1): Query {
+  public static getRawTransaction(txid: string, verbose: number = 0): Query {
```
snowypowers: If you are going to narrow the type for getBlock, then you should be consistent and narrow it for all verbose param.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -108,11 +150,13 @@ export class RPCClient {
...
+   * @param param: `indexOrHash`-height or hash of desired block; `verbose`- 0 for serialized block in hex, 1 for json format, defaults to 0
+   * @returns string or block object according to verbose
    */
-  public async getBlock(
-    indexOrHash: string | number,
-    verbose = 1
-  ): Promise<object | string> {
+  public async getBlock<T extends GetBlockParam>(
+    param: T
+  ): Promise<getBlockResult<T>> {
```
snowypowers: You have mushed up the args in order to get the typing right. Instead of doing this, you should overload the function definitions [instead](https://stackoverflow.com/questions/43958889/return-different-type-by-string-in-typescript).

```
function e(indexOrHash: string, verbose:0): string;
function e(indexOrHash: string): string;
function e(indexOrHash: string, verbose:1): o;
function e(indexOrHash: string, verbose?:0|1): string|o {
    if(verbose === 0) {
        return "hi";
    } else {
        return {h: "hi"}
    }
}
```

The alternative is to ask C# to change their json rpc to return something more consistent. Im of the view that a standard shape request/response structure is a lot easier to work with across languages. We can do this in javascript but it can become a mess in other languages.

packages/neon-core/src/sc/ScriptParser.ts
```diff
@@ -90,6 +90,8 @@ export class ScriptParser extends StringStream {
           output.operation = hexstring2str(output.args!.shift());
           output.args = output.args!.shift();
           return output;
+        case n === 241:
```
snowypowers: comment on what this code stands for.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -275,7 +316,7 @@ export class RPCClient {
   /**
    * This Query returns a list of plugins loaded by the node.
    */
-  public async listPlugins(): Promise<any> {
+  public async listPlugins(): Promise<object[]> {
```
snowypowers: can we type this?

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -20,6 +20,48 @@ export interface GetUnclaimedResult {
   unclaimed: number;
 }
 
+export interface BlockHeader {
```
snowypowers: Because these are using WitnessLike, they should be named with the -Like convention.

packages/neon-core/src/rpc/Query.ts
MorganDream: I think it's better to keep `Query` class to the raw level and only narrow types for RPCClient.

snowypowers: im talking about `verbose: 0|1`

MorganDream: Fixed

hal0x2328:
300 points MorganDream

lllwvlvwlll:

PR #469, opened 2019-09-20
feat(api): TransactionBuilder

MorganDream:
### Design: 
1. TransactionBuilder
*  to build a transaction more conveniently.
* validation
* fee calculation
2. NetProvider
* Deal with read request of neo
3. EntityProvider
* Extends NetProvider
* Use TransactionBuilder and the set account to build transaction and send it.
The design of TransactionBuilder is mainly to do stuffs that are related to transaction but not proper to deal with in Transaction Class.
The design of NetProvider and EntityProvider expects user to use their instances with multiple times. Considering most cases when using sdk:
1. Fixed Network: mainnet or testnet or privatnet, so no need to set url/network as parameter again and again
2. Fixed Sender Accounts: usually user will have one or several accounts in a wallet to invoke transactions, the account param is relatively fixed, no need to set it again and again.
I believe this design is more consistent with use case. And it will improve developing efficiency and give a more clear sight of high level API.
Refer to https://github.com/CityOfZion/neon-js/issues/441

#### TransactionBuilder
```ts
const txBuilder = new TransactionBuilder(url, account);
txBuilder.loadFromConfig({
  cosigners: [{
      account,
      scopes: 0x01
  }]
}).addIntents()
.signWithAquiredAccounts()
.validate(true)
.then(txBuilder => txBuilder.execute());

txBuilder.reset()
.addCosigners()
.addIntents()
.addAttributes()
.signWithAquiredAccounts()
.validate(true)
.then(txBuilder => txBuilder.execute());
```
### API Levels - Entire Workflow
Low Level: neon-core
Middle Level: TransactionBuilder
High Level: NetProvider, EntityProvider

Every Level can be directly invoked to act as sdk.
Higher Level will invoke lower level APIs internally.
The higher the level, the more user-friendly. 
High Level APIs is mostly about direct business functions like `getBalance`, `SendAssets`


snowypowers:
Pls focus on the core features and break out the bug fixes in separate PRs. Before designing the builder APIs I think it is worth an effort to check with users about their needs.

MorganDream:
@snowypowers Now just TransactionBuilder class is left, I can add test when it pass review.
Also I fixed some error and did some code refactoring in neon-core.

MorganDream:
@snowypowers 
1. PR splitted
2. The use case demo and the entire workflow is updated to pr description.
3. From developer side, I think it's better to enable user to do all the things in one place, no need to use functions from different parts, which is kind of annoying.

snowypowers:
- `loadFromConfig` and `reset` are superfluous methods at the moment. I would prefer not to have them. 
- I generally dislike mixing sync and async APIs especially when the fluent pattern is applied.
- I think splitting them into separate components representing different parts of the workflow makes a lot more sense in terms of API. 

```
var builder = new TxBuilder();
var tx = builder.addIntent(sendBobSomeNeo)
.addIntent(sendAliceSomeGas)
.addAttribute(myComment)
.build();

var networkFacade = new NetworkFacade("https://neonode.com:20333");
await networkFacade.calculateFee(tx);

var ledgerSigner = TxSigner.useLedger();
await ledgerSigner.sign(tx);

var success = await networkFacade.sendTransaction(tx);
```

MorganDream:
@snowypowers 
There are 3 times that a transaction need a network facade:
- validUntilBlock needs reference to current height.
- fee calculation
- send tx
Especially the first 2 points, which are necessary. If `txBuilder` class doesn't have network facade, it'll have to query for the network for 3 times. So why not just include a rpc client into `txBuilder`.

I dislike mixing sync and async APIs, too. However setting value for `validUntilBlock` and `fee` must use async API. I pack them into one async API. And async APIs are used after sync API, which is not so bad.

`loadFromConfig` will help:
- to set config in one sentence.
- to set some props by user's will(fee, validUntilBlock)
And `reset` can recycle the builder instance to be used again.
I think they are necessary.

snowypowers:
1. Even with a rpcclient included, it will query the network 3 times because the information is on 3 different endpoints. There isnt a clear advantage other than being in the same object. That is not an advantage.

2. `loadFromConfig` will produce runtime errors when not configured properly but having separate method calls allow us to surface compile time errors. 

> to set some props by user's will(fee, validUntilBlock)
This should be clearly stated in the API with `setFee` instead of being hidden within the `loadFromConfig` method.

3. `new` is cheaper than you setting every field to null. This also introduces a possible bug if a new field is added and not reset correctly by `reset` method.

MorganDream:
> Even with a rpcclient included, it will query the network 3 times because the information is on 3 different endpoints. There isnt a clear advantage other than being in the same object. That is not an advantage.
Yes. But the advantage is that with rpcClient included, it can do the related stuff in a pre-decided/customized way, which is more like high level API.
> About the code example that you give.
```ts
var builder = new TxBuilder();
var tx = builder.addIntent(sendBobSomeNeo)
.addIntent(sendAliceSomeGas)
.addAttribute(myComment)
.build();

var networkFacade = new NetworkFacade("https://neonode.com:20333");
await networkFacade.calculateFee(tx);

var ledgerSigner = TxSigner.useLedger();
await ledgerSigner.sign(tx);

var success = await networkFacade.sendTransaction(tx);
```
It seems that the only one difference between `txBuilder` and `Transaction` constructor is the API `addIntent`, so is it worth it to open another class to just add one function?

My opinion for `txBuilder` is more like high level api, so it should have some integration APIs in it.

snowypowers:
That is just a simple example. I also intend to:
- replace the `addAttribute` on Transaction to be take a list of TransactionAttributeLike objects. The old `addAttribute` can be on the builder.
- Remove the `sign` method on Transaction class. This should remove the link between tx and wallet folders.

Im not opposed to having RPC methods and other features if done correctly. However,

- Currently, there is only a single constructor requiring a RPC client. That is not acceptable if we want to have the builder work in an offline context.

- The validation logic is coupled very tightly in this implementation. More controls are required as to which parameter can be auto fixed. Auto fixing validUntilBlock isnt a very big deal but autofixing fees might require a pop up to the user before sending. The current `validate` does not allow for this.

- Similarly, there are a multitude of signing scenarios that we have to take care of. Signing by private key, adding multisig, using a ledger are the 3 basic scenarios that all require a different way of expression.

MorganDream:
> Currently, there is only a single constructor requiring a RPC client. That is not acceptable if we want to have the builder work in an offline context.

How about using class `Transaction` as offline builder and `txBuilder` as a stronger and customized online builder? 
> The validation logic is coupled very tightly in this implementation. More controls are required as to which parameter can be auto fixed. Auto fixing validUntilBlock isnt a very big deal but autofixing fees might require a pop up to the user before sending. The current validate does not allow for this.

Well the validations for fees, validUntilBlock, signatures can be separated and selected by user, so it's not coupled. 
```ts
txBuilder.validateValidUntilBlock(true).then(txBuilder => TxBuilder.validateSystemFee())
```
Maybe we should delete API `validate()` to avoid abusing?
About fee autofixing confirmation, user could set `autoFix: false` or just neglect it. And we could add the minimal fee needed in reject params to let user know the minimal fees. About popup to show users, that's what a product developer should consider. And with such API, a product developer is able to handle this.

> Similarly, there are a multitude of signing scenarios that we have to take care of. Signing by private key, adding multisig, using a ledger are the 3 basic scenarios that all require a different way of expression.

Yeah. And is neon-js supporting ledger now?

snowypowers:
> How about using class Transaction as offline builder and txBuilder as a stronger and customized online builder?

The Transaction object will not have the same methods as the TxBuilder so we cannot compare it as offline vs online builders. This suggestion is kinda breaking the abstraction layer too.

Well the validations for fees, validUntilBlock, signatures can be separated and selected by user, so it's not coupled.

Ok good. I feel that the class is just reaching a tipping point in terms of size and there is benefits is pulling it into a separate class for network based validation.

> Yeah. And is neon-js supporting ledger now?

The old API allowed for ledger signing. Without a firmware update it is impossible to sign for neo3 but preparing for remote signing should be on the table.

MorganDream:
@snowypowers 
1. validation is taken out, and rpcClient is not within builder
2. modify addAttribute
3. consider multisig account

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Cosigners
+   */
+  private _cosigners: CosignerLike[] = [];
+  /**
+   * Cosigner accounts that client has access to
+   */
+  private _cosignerAccounts: Account[] = [];
+  private _intents: ScriptIntent[] = [];
+  private _transaction: Transaction;
```
snowypowers: Having the transaction field and also the other fields might be confusing. Consider either having only the transaction field or just the intermediate fields.

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+      Object.assign(config, {
+        sender: reverseHex(this._sender.scriptHash),
+        cosigners: this._cosigners
+      })
+    );
+    this.addCosigners(...cosigners);
+    this.addIntents(...intents);
+  }
+
+  public get transaction() {
```
snowypowers: There isnt a need for this api.

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+  private _cosignerAccounts: Account[] = [];
+  private _intents: ScriptIntent[] = [];
+  private _transaction: Transaction;
+
+  /**
+   * A TransactionBuilder instance can build class with the same network and sender account
+   * @param rpc to specify network, can be RPCClient or rpc url
+   * @param sender the sender account, can be Account or privateKey
+   */
+  public constructor(config: TransactionBuilderConfig) {
```
snowypowers: consider a design where the constructor is empty and we do import/export using a normal Transaction object.

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+    this._transaction.sign(account);
+    return this;
+  }
+
+  /**
+   * The transaction must be signed with sender and all other cosigners.
+   * It may be that all signatures cannot be aquired from one client.
+   * This function is to sign the transaction with (sender) aquired accounts.
+   */
+  public signWithAquiredAccounts(): this {
```
snowypowers: what is aquiredAccounts.

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  public setSystemFee(fee: Fixed8) {
+    this._transaction.systemFee = fee;
+  }
+
+  public setNetworkFee(fee: Fixed8) {
+    this._transaction.networkFee = fee;
+  }
+
+  public export(): Transaction {
```
snowypowers: use `build()` and it should produce a new transaction, not just return the reference to the private field.

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+
+    const multisigWitness = Witness.buildMultiSig(
+      this._transaction.serialize(),
+      witnesses,
+      multisigAccount
+    );
+    return this.addWitnesses(multisigWitness);
+  }
+
+  public sign(account: Account | string): this {
```
snowypowers: remove

packages/neon-api/src/txService/builder.ts
```diff
@@ -0,0 +1,186 @@
...
+  networkFee?: number;
+  validUntilBlock?: number;
+  attributes?: TransactionAttributeLike[];
+  scripts?: WitnessLike[];
+  script?: string;
+}
+
+/**
+ * `TransactionBuilder` is to build a transaction in a more convinient and mature way.
+ * It is more powerf
```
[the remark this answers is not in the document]
MorganDream: I'll do this and add ut when we settle the primary codes

packages/neon-api/src/txService/builder.ts
MorganDream: I use the builder style and removed transaction.

packages/neon-api/src/txService/builder.ts
MorganDream: I am not clear about this. Could you explain more?

packages/neon-api/src/txService/builder.ts
MorganDream: Removed

MorganDream:
@snowypowers I agree with your opinion on `signer` and `validator`.
> Add a estimateFees(priorityFee) which will just set the fees as the minimum required given the current data without a network request. This might be useful for a wallet if they want to suggest a fee to the user.

Will this be useful? Cause the offline minimum required fee calculation is not accurate.
> We might want to create a new Intent class to go with this since NEP5 intents will become the focus of this class. I imagine this new Intent class alone might be sufficient to create a transaction by itself.

I already opened a PR https://github.com/CityOfZion/neon-js/pull/459 about this intent classes.
However I don't think it's proper to use such class to create a transaction, because a transaction may contain multiple intents. Using intent to make transaction is not scalable design.
I just use it to generate script.

packages/neon-api/src/index.ts
```diff
@@ -21,10 +21,14 @@ function assignSettings(
 }
 function bundle<T extends typeof _Neon>(
   neonCore: T
-): T & { api: typeof plugin } {
+): T & {
+  TransactionBuilder: TransactionBuilder;
```
snowypowers: if we are using the `api` package, then all these should be under the `api` namespace.

packages/neon-api/src/transaction/util.ts
```diff
@@ -0,0 +1,84 @@
+import { tx, wallet, u, sc, CONST } from "@cityofzion/neon-core";
```
snowypowers: Do we still need this for this PR?

packages/neon-core/src/tx/transaction/Transaction.ts
```diff
@@ -197,13 +197,8 @@ export class Transaction {
...
   }
 
-  /**
-   * Add an attribute.
-   * @param usage The usage type. Do refer to txAttrUsage enum values for all available options.
-   * @param data The data as hexstring.
-   */
-  public addAttribute(usage: number | string, data: string): this {
-    this.attributes.push(new TransactionAttribute({ usage, data }));
+  public addAttribute(...attributes: Array<TransactionAttribute>) {
```
snowypowers: lets standardise if we want to accept a TransactionAttribute, TransactionAttributeLike,  Array<TransactionAttribute> or Array<TransactionAttributeLike> here. All the methods currently have a different style. I prefer adding a single TransactionAttributeLike.

packages/neon-api/__tests__/transaction/builder.ts
```diff
@@ -0,0 +1,60 @@
...
+
+describe("setter", () => {
+  test("addAttributes", () => {
+    const transaction = new TransactionBuilder()
+      .addAttributes({
+        usage: 129,
+        data: "abcd"
+      })
+      .build();
+    expect(transaction.attributes[0].data).toBe("abcd");
```
snowypowers: please use the equality method for all the tests instead of comparing only a single field. Also test adding multiple items correctly.

packages/neon-api/src/transaction/util.ts
MorganDream: Removed

packages/neon-core/src/tx/transaction/Transaction.ts
snowypowers: I dont really mind having the spread param but you will have to update the method name to use the plural form.

MorganDream: I have changed all add methods to accept `Spread Array` and `Like`, I don't think the method name should be plural, as spread param method name should not be plural

snowypowers: The methods on the builder are all spread params and in plural form. 

My argument against spread params is that it will be difficult to adjust the api in the future if there is a need to introduce or tweak an argument. The spread param also requires annotations on the API itself. It is not clear that the method provides the capability to add multiple items if the name is singular.

MorganDream: OK. Base class interfaces better be simple.
I changed to single param and removed unnecessary logic for `addWitness`, please check.

hal0x2328:
400 points MorganDream

lllwvlvwlll:

PR #479, opened 2019-10-08
fix(wallet): Fix wallet tests and error handling

snowypowers:
We have to update the test data when transferring to the neo3 because NEP2 uses a hash of the address and the address in neo3 has changed due to the change in scripthash.

- Update wallet tests to use async await style
- Update wallet json in test
- wallet encrypt decrypt to throw instead of returning Promise.reject. This was causing jest not to catch the errors due to double wrapped promises.


hal0x2328:
200 points snowypowers

lllwvlvwlll:

Issue #480, opened 2019-10-10
Proper hexstring support


snowypowers:
There can be improved support internally for hexstrings to differentiate between endianness, resolve the `0x` sign and support for typing of different hexstrings.

snowypowers:
This issue is primarily for neo3 but it is also a consideration for porting to neo2. However, I would consider it low priority given the potential work and mess.

MorganDream:
I have proposed a PR and want to have discussion with you on below questions:
1. Where should we put these types, and what's the relation between these types and old type-converting apis.
Suggestion: We put these types into module `u`, and stop exporting old type-conversion apis like `str2hextring` and include them into these types, to avoid confusions.

2. `-Like` interfaces
The `-Like` interfaces (e.g. CosignerLike) mostly use string as type, and they will be used to type param for constructors. So shall change the `-Like` interfaces to use new types.
Currently I create a new interface `-Config` to act as the param for constructor.

snowypowers:
1. Yes the `u` module would be the correct place for these. I would bring in the existing methods and unify them under the common namespace.

2. The `-Like` interfaces have been used to represent the json counterpart that we are exporting. It acts as the gateway between the outside world and the neon objects. The constructor must be able to take in a json object and transform it into an neon object. Please continue using the `-Like` interface.

MorganDream:
I think one of the biggest advantages using strong types is that user can use these types when writing codes to avoid confusion. And constructors are used a lot in neon-js. If only `string` is used to initiate some class, the generated instances would probably be invalid or we need to add redundant logic to throw error (e.g. a scripthash param would expect a 30-length string, user may input a "" to be a scripthash, or not even a hex).

Suggestion: 
So I think it's better to type the params for constructors.
Add a static function like `fromJson` to accept the `-Like` param.

snowypowers:
For this to happen, all neon-js classes will have to migrate to the new pattern. Let us see how the Hexstring class works out first before committing.

PR #482, opened 2019-10-14
feat(u): Hexstring class

MorganDream:
This is tryout for solution to https://github.com/CityOfZion/neon-js/issues/480.
I have created below types:
- HexString
- ScriptHash
- PublicKey
- Address
And I have tried to use them in Class Cosigner and TransactionAttribute.

And Let's talk about several design issues about this feature:
1. Where should we put these types, and what's the relation between these types and old type-converting apis.
Suggestion: We put these types into module `u`, and stop exporting old type-conversion apis like `str2hextring` and include them into these types, to avoid confusions.

2. `-Like` interfaces
The `-Like` interfaces (e.g. CosignerLike) mostly use string as type, and they will be used to type param for constructors. So shall change the `-Like` interfaces to use new types.
Currently I create a new interface `-Config` to act as the param for constructor.


snowypowers:
Please focus on the hexstring class first. The others can be filled in later.

MorganDream:
Now only HexString class is left. And let's settle this first.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+    this._value = value;
+    this._littleEndian = littleEndian;
+  }
+
+  /**
+   * Get hex value as string
+   * @param asLittleEndian true, export a little-endian string; false, export a big-endian string; default to be false
+   * @returns string
+   */
+  public value(asLittleEndian = false): string {
```
snowypowers: replace `value` and `reverse` with `toLittleEndian` and `toBigEndian`. Alternatively, we can use `toString` and `toReverseString` as the API instead.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  public toNumber(): number {
+    return parseInt(this.value(), 16);
+  }
+
+  public toArrayBuffer(): Uint8Array {
+    return hexstring2ab(this.value());
+  }
+
+  public static fromHexString(str: string, littleEndian = false) {
```
snowypowers: fromHex

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+    return new HexString(str, littleEndian);
+  }
+
+  public static fromASCIIString(str: string, littleEndian = false) {
```
snowypowers: fromAscii(). And remove the littleEndian argument.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+    const hex = str2hexstring(str);
+    return new HexString(littleEndian ? reverseHex(hex) : hex, littleEndian);
+  }
+
+  public static fromNumber(num: number, littleEndian = false) {
+    return new HexString(num2hexstring(num, 1, littleEndian), littleEndian);
+  }
+
+  public static fromArrayBuffer(arr: ArrayBuffer | ArrayLike<number>) {
```
snowypowers: I think arraybuffers also have an endianness.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+    ensureHex(value);
+  }
+
+  protected constructor(value: string, littleEndian = false) {
+    if (value.startsWith("0x")) {
+      value = value.slice(2);
+    }
+    this._checkValue(value);
+    this._value = value;
+    this._littleEndian = littleEndian;
```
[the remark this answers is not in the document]
MorganDream: Yeah, good idea.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  public isEqualTo(other: HexString): boolean {
+    return this.value() === other.value();
+  }
+
+  public xor(other: HexString): HexString {
+    return HexString.fromHexString(hexXor(this.value(), other.value()));
+  }
+
+  public toASCIIString(): string {
```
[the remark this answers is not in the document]
MorganDream: Changed

packages/neon-core/src/u/HexString.ts
MorganDream: The `value` can be replaced with `toLittleEndian` and `toBigEndian` but not `reverse`.
`reverse` is a write function to change the value of HexString.

I used `export(asLittleEndian: boolean)` to replace `value` cause I think `toLittleEndian` and `toBigEndian` sounds like they're going to do some conversion to change the value.

packages/neon-core/src/u/HexString.ts
MorganDream: I also removed littleEndian param for `fromNum`

packages/neon-core/src/u/HexString.ts
snowypowers: the user will have to specify the endianness on creation. Its similar to `fromHex`.

packages/neon-core/src/u/HexString.ts
snowypowers: Remove the write function and we have an immutable class that is very clear in terms of API. If we need to reverse the internal value, it means that we stored the value wrongly. There should never be a need to reverse the value.

We should always be stating how we are using the hexstring values so `toLittleEndian` and `toBigEndian` works nicely.

packages/neon-core/src/u/HexString.ts
snowypowers: this is slightly debatable as a number representation can be in either endian.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  public toBigEndian(): string {
+    return this._value;
+  }
+
+  public toLittleEndian(): string {
+    return reverseHex(this._value);
+  }
+
+  public reverse(): this {
```
snowypowers: please remove this. There should be no cases where this comes in useful.

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,80 @@
...
+    this._value = reverseHex(this._value);
+    return this;
+  }
+
+  public isEqualTo(other: HexString): boolean {
```
snowypowers: we use `equals` on all our other classes so stick with the convention.

packages/neon-core/src/u/HexString.ts
MorganDream: OK

packages/neon-core/src/u/HexString.ts
MorganDream: OK

packages/neon-core/src/u/HexString.ts
```diff
@@ -0,0 +1,122 @@
...
+   * Export as ASCII string
+   */
+  public toAscii(): string {
+    return hexstring2str(this.toBigEndian());
+  }
+
+  /**
+   * Export as number
+   */
+  public toNumber(): number {
```
snowypowers: if we have the option to import with an custom endian option, we should allow export with the same custom option. Similar for arraybuffers.

packages/neon-core/__tests__/u/HexString.ts
```diff
@@ -0,0 +1,70 @@
...
+    const hex = HexString.fromHex("a4c2");
+    const arr = hex.toArrayBuffer();
+    expect(arr[0]).toBe(164);
+    expect(arr[1]).toBe(194);
+  });
+});
+
+describe("math operation", () => {
+  test("xor", () => {
+    const hex = HexString.fromHex("a4c2");
```
snowypowers: pls add an comment describing the bits so its easier to see the effects

```
const hex = HexString.fromHex("a4c2"); //1010 0100 1100 0010
```

MorganDream: All modified.

hal0x2328:
300 points morgandream

lllwvlvwlll:

PR #485, opened 2019-11-12
feat(api): TransactionSigner

MorganDream:
A class used to sign transaction:
- Sign with account or private key.
- Sign with witnesses
- Sign with multi-sig


packages/neon-api/src/transaction/util.ts
```diff
@@ -0,0 +1,21 @@
+import { tx, wallet, u } from "@cityofzion/neon-core";
+
+export function checkMultisigAcc(
```
snowypowers: this looks very specific to the multisig method. I would just put it into the signer class and private it. It doesnt look like it has a lot of use outside.

MorganDream:
> It looks like the multi sig is the only method that checks if the cosigners/sender are part of the transaction. Are the rest of the methods not required to check that condition?

Let's standardize it. I propose remove checking this, because:
- User can add sender or cosigner after signing.
- We can enable Transaction validator to check this.

snowypowers:
Would that make sense? Adding sender/cosigners will invalidate any previous signatures attached and the whole process will have to repeat.

Another question is will adding additional signatures fail the transaction?

MorganDream:
> Would that make sense?

It will make sense.  Check if signature-sender/cosigner is paired is the validation work. We should enable user to sign a transaction if he want that.
>  Adding sender/cosigners will invalidate any previous signatures attached and the whole process will have to repeat.
Why need to repeat?  Transaction signing should be done before validation. 
We should allow user to do some configuration in a flexible way, User can just `addCosigner`.

> Another question is will adding additional signatures fail the transaction?
It will

snowypowers:
Cosigner is included as part of the transaction serialization process so adding a cosigner will change the transaction hash which in turns invalidates the old signature.

packages/neon-api/src/transaction/util.ts
MorganDream: OK

MorganDream:
@snowypowers Now checking account is also added for `signWithAccount`, but it seems hard to check witness.

snowypowers:
The witness contains the public key which can be transformed to the scripthashes.

MorganDream:
CheckWitness Added

packages/neon-api/src/transaction/signer.ts
```diff
@@ -0,0 +1,103 @@
...
+      this.transaction.addWitness(witness);
+    });
+  }
+
+  /**
+   * Sign a transaction with multi-signatures for multi-sig account
+   * @param multisigAccount multisig account
+   * @param witnesses signatures from accounts within the multisig-account
+   */
+  public addMultiSig(
```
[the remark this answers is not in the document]
MorganDream: OK

packages/neon-api/src/transaction/signer.ts
```diff
@@ -0,0 +1,103 @@
...
+      );
+    } else if (publicKeys.length === 1) {
+      this._checkScripthash(wallet.getScriptHashFromPublicKey(publicKeys[0]));
+    } else {
+      throw new Error("Invalid witness: Invalid verification script.");
+    }
+  }
+
+  private _checkMultisigAcc(multisigAcc: wallet.Account) {
+    if (!multisigAcc.isMultiSig) {
```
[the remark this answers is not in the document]
MorganDream: I don't get it, do you mean `isMultisig` will be false when `contract.script = ""` but actually this account is a multisig account?
`contract.script` is needed in the following `buildMultisig` process, so it's OK to throw at such case.
I changed the throwing error info to get it clear.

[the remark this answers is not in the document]
snowypowers: There is  multiple ways to construct an `Account` object so the `contract` field is not guaranteed to be populated correctly. However i think it should be okay in this case where they should have the full details.

packages/neon-api/src/transaction/signer.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.transaction.addWitness(multisigWitness);
+  }
+
+  private _checkAcc(account: wallet.Account | string) {
+    const acc = new wallet.Account(account);
+    this._checkScripthash(acc.scriptHash);
+  }
+
+  private _checkWitness(witness: tx.Witness) {
+    const publicKeys = wallet.getPublicKeysFromVerificationScript(
```
snowypowers: I dont understand all this. Why go through so much when you can calculate the scripthash from the verification script directly in one step? Its the same scenario for both single and multisig accounts.

packages/neon-api/src/transaction/signer.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  private _getSignerHashes() {
+    return [
+      this.transaction.sender,
+      ...this.transaction.cosigners.map(cosigner => cosigner.account)
+    ].map(hash => u.reverseHex(hash));
+  }
+
+  private _checkScripthash(scriptHash: string) {
```
snowypowers: Consider renaming to `assertShouldSign` or something to indicate what we are checking for. Currently the name just suggests that we check if the string looks like a scripthash.

snowypowers:
I noticed that there is an increasing number of lint warnings. Please make sure that your changes do not have any lint warnings. The base branch now includes logic to fail the lint check if the changed files have any lint warnings.

hal0x2328:
300 points MorganDream

lllwvlvwlll:

PR #486, opened 2019-11-12
feat(api): TransactionValidator

MorganDream:
Transaction Validator for Neo3, can validate:
- validUntilBlock
- systemFee
- networkFee
- signatures

Validation will return ValidationResult for developer to handle.


MorganDream:
> For the validator, let me propose an alternative design. From a wallet's perspective, I imagine the usage to be:

> Craft transaction with initial user intents and initial fee estimates.
Perform validation and display the changes to the user.
User agrees and proceeds to signing.
To be inline with this, we should allow a codeflow that can surface the changes and then allow the system to accept the appropriate changes. Instead of rejecting the promises when there are changes, we should return the changes instead.

@snowypowers I quoted your suggestion on previous PR and let's discuss it here.
Validation of transaction is not only about fee, it can also validate intents and signatures. If intents and signatures is modified after validation, the previous validation for fees should be outdated. It's not applicable in this case to apply those changes.

Please check my new design. I suggest not to provide functions to apply validation suggestions, but just return it to developers and let developers handle this.

snowypowers:
1. It is weird that your individual functions return `ValidationResult` but only one of the fields will be filled.

2. Because signatures affect the cost of the transaction, I was considering adding unsigned witnesses during the builder phase so we can calculate the cost effectively.

3. I suggested a combined way in order to allow accepting suggestions. Im okay with developers handling the suggestions themselves.

4. I would like to be consistent and return a consistent result structure from all validation methods. It should contain a machine readable suggestion and also a human readable message.

MorganDream:
> 1. It is weird that your individual functions return `ValidationResult` but only one of the fields will be filled.

I haven't added the `validateAll` function, so user could do several validations at one time, then the result will contains several fields. Individual function returns the same type is easy for them to combine to one result.
> 2. Because signatures affect the cost of the transaction, I was considering adding unsigned witnesses during the builder phase so we can calculate the cost effectively.

Why not validate the transaction after user has signed it? or you mean to calculate in builder?
> 4. I would like to be consistent and return a consistent result structure from all validation methods. It should contain a machine readable suggestion and also a human readable message.

`ValidationResult` is a consistent result type. And it now has a machine readable message.
Well I agree to add a human readable message.

snowypowers:
> I haven't added the `validateAll` function, so user could do several validations at one time, then the result will contains several fields. Individual function returns the same type is easy for them to combine to one result.

How are you going to combine `result` and `fixed` into one result? On a side note, what does `result` even mean? I think it would be better to return `ValidationSuggestion` for each validation function if we are doing each one individually. How are you going to let the user control `validateAll`? 

> Why not validate the transaction after user has signed it? or you mean to calculate in builder?

Because at this stage there isnt much use of validation. Any changes to the transaction will cause a resign so might as well just send the transaction to the network.

If we add unsigned witnesses at the builder stage, we can calculate the signing fees correctly and produce quite a good fee estimate.

MorganDream:
> How are you going to combine result and fixed into one result? On a side note, what does result even mean? 

`result` means whether the transaction is valid in current validation.
`fixed` means whether the transaction has been modified by validator to make it valid. This can be TRUE when user specify `autoFix` as TRUE. But there are cases validator cannot fix like signing. `fixed` doesn't mean the transaction is fixed to be valid, but modified to achieve validation.
```ts
public async validateAll(autoFix = false) {
   return Promise.all([this.validateSystemFee()...]).then(res => {
     result: AND between all result,
     fixed: OR between all fixed
   })
}
```
Yeah, I can see the problem may be user don't know which validation is fixed which is not.
So how about moving fixed into validationSuggestion? Result can tell user whether transaction is valid after `validateAll`, `suggestions` can tell them the suggestion and what is fixed.

> Because at this stage there isnt much use of validation. Any changes to the transaction will cause a resign so might as well just send the transaction to the network.

> If we add unsigned witnesses at the builder stage, we can calculate the signing fees correctly and produce quite a good fee estimate.
Well as far as I can see, I think fee estimation is useless. As a user, I always want to spend as little as possible gas to complete my transaction, fee estimation will probably be invalid or make me pay more, which make it feel useless.

MorganDream:
A better and more scalable design may be that:
```ts
type ValidationAttr = 'systemFee' | 'networkFee' | 'witness';
interface ValidationSuggestion<T> {
  attr: ValidationAttr,
  prev: T,
  suggestion: T
}
public async validate(attrs: Array<ValidationAttr>, autoFix: boolean);

```

snowypowers:
Yes I would think that `ValidationResult` simply be `{ [transactionProperty: string]: ValidationSuggestion }`. I prefer using a map over an array as order wouldnt matter.

`result` field is kinda useless cos if we are giving a suggestion, it means its not valid.

Im okay with the `validate` design. Please use flags for the type, it will be nicer.

MorganDream:
> Yes I would think that ValidationResult simply be { [transactionProperty: string]: ValidationSuggestion }
There are some cases that we cannot fix and give the same type suggestion: Signing.
Or should we remove it?

> result field is kinda useless cos if we are giving a suggestion, it means its not valid.
If user set `autoFix: true` and validator fixes all to make it valid, we should give the ValidationSuggestion to tell them it's changed from before validation, and result is a fast property to tell them the transaction is valid now.

In newest code, I have change `result` to `valid`

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+  //     result: true,
+  //     fixed: false
+  //   };
+  // };
+
+  /**
+   * validate systemFee
+   * @param autoFix will automatically fix transaction if specified as true
+   */
+  validateSystemFee = async (autoFix = false): Promise<ValidationResult> => {
```
snowypowers: 1. You might want to do a check on the existing systemFee to make sure that it is in an acceptable format too (integer without decimals).

2. Consider including the action to verify the intent here too. If the intent is invalid, the systemFee validation will also probably be worthless.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+        valid: false,
+        networkFee: {
+          fixed: false,
+          prev: networkFee,
+          suggestion: minimumNetworkFee
+        }
+      };
+    }
+    return {
+      valid: true
```
snowypowers: So there isnt a suggestion if we do not want autofix but attached a larger than needed fee?

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+          suggestion: vub_suggestion
+        }
+      };
+    }
+    return {
+      valid: true
+    };
+  };
+
+  // validateIntents = async (): Promise<ValidationResult> => {
```
snowypowers: Remove comment

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+  /**
+   * Whether this is auto-fixed by validator
+   * Validator will try to fix the transaction when param `autoFix` is set to TRUE
+   */
+  fixed: boolean;
+  prev: T;
+  suggestion: T;
+}
+
+export interface ValidationResult {
```
snowypowers: Consider 
```
{ 
  valid: boolean,
  suggestions: {[prop: string] : ValidationSuggestion<T>
} where prop in Transaction
```

If there is no way to suggest a fix, leave suggestion `null` but offer a human readable message on the issue.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+      valid: false,
+      witnesses: {
+        notSignedHash
+      }
+    };
+  };
+
+  async validate(
+    attrs: ValidationAttributes,
+    autoFix = false
```
snowypowers: can we also have `autoFix` be `ValidationAttributes` so we can select properties to autofix?

packages/neon-api/src/transaction/util.ts
```diff
@@ -0,0 +1,64 @@
...
+    })
+  );
+}
+
+export function getNetworkFee(transaction: tx.Transaction): u.Fixed8 {
+  const networkFee = new u.Fixed8(0);
+  const signers = transaction.getScriptHashesForVerifying();
+  signers.forEach(signer => {
+    const account = new wallet.Account(signer);
+    if (!account.isMultiSig) {
```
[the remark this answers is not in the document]
MorganDream: Now I calculate fee from signatures

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+    return {
+      valid: true
+    };
+  };
+
+  /**
+   * Validate Signatures
+   */
+  validateSigning = async (): Promise<ValidationResult> => {
```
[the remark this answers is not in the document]
MorganDream: Signatures are part of transaction and signing should be done before validation I think.

packages/neon-api/src/transaction/validator.ts
MorganDream: 1. Yeah, thanks for reminding.
2. OK, will do. Planning to using manifest

packages/neon-api/src/transaction/validator.ts
MorganDream: Yes. Anyway validator is mostly to validate, to save money is secondary job. But it's harmless if we add a check here, what do you think?

packages/neon-api/src/transaction/validator.ts
MorganDream: Cannot use `[prop: string] : ValidationSuggestion<T>` cause we need to pre-define `T`.
now it's 
```ts
suggestions?: {
    validaUntilBlock?: ValidationSuggestion<number>;
    systemFee?: ValidationSuggestion<u.Fixed8>;
    networkFee?: ValidationSuggestion<u.Fixed8>;
    witnesses?: string;
  };
```

packages/neon-api/src/transaction/validator.ts
MorganDream: Emm.. then the flag thing should be changed, right?
I believe user can invoke twice, one with autoFix, one without to achieve this scope.

packages/neon-api/src/transaction/validator.ts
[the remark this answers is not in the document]
snowypowers: I thought we have discussed this multiple times. The order should be build, validate and sign. Signing must always be the last step because any adjustments during the build and validate step will require a resigning.

packages/neon-api/src/transaction/validator.ts
snowypowers: Im not clear if in neo3 the VM will error out on scripts requiring verification. If it does, then it might be hard to validate the script and estimate the fee.

packages/neon-api/src/transaction/validator.ts
snowypowers: If you say that, then autofix should not fix if the attached fee is larger than required.

packages/neon-api/src/transaction/validator.ts
snowypowers: Can we have the suggestion reflect the properties of Transaction? So instead of `witnesses` its `scripts`.

packages/neon-api/src/transaction/validator.ts
snowypowers: > Emm.. then the flag thing should be changed, right?

What do you mean? 

> I believe user can invoke twice, one with autoFix, one without to achieve this scope

Yes they could. I think the autoFix change is a small natural improvement using the same set of flags.

MorganDream: Oh, I see it. Will do.

packages/neon-api/src/transaction/validator.ts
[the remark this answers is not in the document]
MorganDream: I have removed validateSigning, however how can we add such empty witnesses?

packages/neon-api/src/transaction/validator.ts
MorganDream: Not very clear about your concern, I have added `validateIntents` using ContractManifest.

packages/neon-api/src/transaction/validator.ts
MorganDream: `autoFix: ValidationAttributes | boolean`
Give user the freedom to assign one boolean, too.

packages/neon-api/src/transaction/validator.ts
MorganDream: Well I think autoFix could fix when attached fee is larger than required.
It will not give suggestions when autoFix = false and attached fee is larger than required.
What do you think?

packages/neon-api/src/transaction/validator.ts
snowypowers: you can create the convenient flags `ValidationAttributes.ALL` and  `ValidationAttributes.NONE`

packages/neon-api/src/transaction/validator.ts
snowypowers: I think most people will expect the `autoFix` flag to just apply the suggestion and not change the suggestion output based on that flag. Please make it consistent.

MorganDream: OK.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    if (operation) {
+      if (
+        manifest.abi.methods.map(method => method.name).indexOf(operation) < 0
+      ) {
+        return `Unknow method ${operation} for constract ${scriptHash}`;
+      }
+    }
+
+    return undefined;
```
snowypowers: You might want to consider returning empty string or null instead of undefined. Undefined is usually not intentional.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,298 @@
...
+        this._validateIntent(intent).then(res =>
+          res ? intentsRes.push(res) : res
+        )
+      )
+    );
+    if (intentsRes.length > 0) {
+      return {
+        valid: false,
+        suggestions: {
+          script: intentsRes
```
snowypowers: We should return a ValidationSuggestion and add a `message` field that indicates a human readable error. At this point, you might want to not add the `message` feature first and complete this PR.

packages/neon-core/src/rpc/RPCClient.ts
```diff
@@ -220,10 +221,14 @@ export class RPCClient {
...
    * Gets the state of the contract at the given scriptHash.
    */
-  // TODO: parse result to manifest when contract manifest is merged
-  public async getContractState(scriptHash: string): Promise<object> {
+  public async getContractState(
+    scriptHash: string
+  ): Promise<ContractManifest | null> {
     const response = await this.execute(Query.getContractState(scriptHash));
-    return response.result;
+    if (response.error) {
```
snowypowers: the rpc client will throw if there is an error. I dont think hiding the error is a good idea anyway.

MorganDream: If we just catch the throwing error, we cannot tell exactly whether the error is because of network error or contract not exist.

If contract doesn't exist, it will return 200, too, with an error info in the response body. We can firmly tell user it doesn't exist.

packages/neon-api/src/transaction/validator.ts
MorganDream: I just added this. Please check

packages/neon-core/src/rpc/RPCClient.ts
snowypowers: Ah i see my bad I misread the code. Maybe time to rework the RPC client a little.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,303 @@
...
+      }
+    }
+
+    return null;
+  }
+
+  /**
+   * Validate intents
+   */
+  async validateScript(): Promise<ValidationResult> {
```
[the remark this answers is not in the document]
MorganDream: Indeed.
Something that need to be considered: 
1. If we add online testing for script in validateScript and user validate both systemFee and script, the same request would be sent twice.
2. At the begging of testing on neo3 preview testnet, a script will return 'FAULT' when `invokeScript` but actually works in a transaction. I'll try to reproduce this problem and ask someone else if it is a bug.

MorganDream:
I'll add test later.

MorganDream:
@snowypowers Now ut is added

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,314 @@
...
+      fixed: boolean;
+      prev: T;
+      suggestion: T;
+      message?: T;
+    }
+  | {
+      fixed?: boolean;
+      prev?: T;
+      suggestion?: T;
+      message: T;
```
snowypowers: `message` field should be string.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,314 @@
...
+      /**
+       * Whether this is auto-fixed by validator
+       * Validator will try to fix the transaction when param `autoFix` is set to TRUE
+       */
+      fixed: boolean;
+      prev: T;
+      suggestion: T;
+      message?: T;
+    }
+  | {
```
snowypowers: Is there any way we can simplify this? Do we have to make `fixed` optional or can we just always return something (`false` if `autoFix` is false?).

packages/neon-api/__tests__/transaction/validator.ts
```diff
@@ -0,0 +1,320 @@
...
+  verificationScript:
+    "5221031d8e1630ce640966967bc6d95223d21f44304133003140c3b52004dc981349c92103767002bb9f74317035ce8d557a3aed30ce831eb16b5f636a139dad0b07916bed210329898e6e5e0a2f175e205b4019c500d6bb69203b56470ec2fc8ab0a4c065e16d5368c7c34cba"
+};
+
+describe("validateValidUntilBlock", () => {
+  test("valid", async () => {
+    const transaction = new TransactionBuilder({
+      validUntilBlock:
+        tx.Transaction.MAX_TRANSACTION_LIFESPAN +
+        (await rpcClient.getBlockCount()) -
```
snowypowers: I think we can move the entire suite to the integration folder instead.

packages/neon-api/__tests__/transaction/validator.ts
```diff
@@ -0,0 +1,320 @@
...
+    expect(result.valid).toBeTruthy();
+  });
+
+  test("invalid", async () => {
+    const transaction = new TransactionBuilder({
+      validUntilBlock: 1
+    }).build();
+    const validator = new TransactionValidator(rpcClient, transaction);
+    const result = await validator.validateValidUntilBlock();
+    expect(result.valid).toBeFalsy();
```
snowypowers: Not checking the other fields?

packages/neon-api/__tests__/transaction/validator.ts
```diff
@@ -0,0 +1,320 @@
...
+    });
+    const transaction = new TransactionBuilder({
+      script
+    }).build();
+    const validator = new TransactionValidator(rpcClient, transaction);
+    const result = await validator.validateScript();
+    expect(result.valid).toBeTruthy();
+  });
+
+  test("inValid", async () => {
```
snowypowers: typo

packages/neon-api/__tests__/transaction/validator.ts
```diff
@@ -0,0 +1,320 @@
...
+    const transaction = new TransactionBuilder({
+      scripts: [sig, multiSig],
+      networkFee: 0.0237654
+    }).build();
+    const validator = new TransactionValidator(rpcClient, transaction);
+    const result = await validator.validateNetworkFee();
+    expect(result.valid).toBeTruthy();
+  });
+
+  test("inValid", async () => {
```
snowypowers: typo

packages/neon-api/__tests__/transaction/validator.ts
```diff
@@ -0,0 +1,320 @@
...
+      networkFee: 0.01,
+      script,
+      systemFee: 1.1
+    }).build();
+    const validator = new TransactionValidator(rpcClient, transaction);
+    const result = await validator.validate(ValidationAttributes.All);
+    expect(result.valid).toBeFalsy();
+  });
+
+  test("autoFix", async () => {
```
snowypowers: add test for validating all but autofixing only 1.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,314 @@
...
+    if (!validateIntentsResult.valid) {
+      return validateIntentsResult;
+    }
+
+    const { script, systemFee } = this.transaction;
+    const { gas_consumed: gasConsumed } = await this.rpcClient.invokeScript(
+      script
+    );
+    const minimumSystemFee = new u.Fixed8(
+      parseFloat(gasConsumed) * 1e-8
```
snowypowers: please dont do multiply outside of Fixed8.

packages/neon-api/__tests__/transaction/validator.ts
MorganDream: Check the whole object when autoFix = true. Here we only care about whether it's valid

packages/neon-api/src/transaction/validator.ts
MorganDream: Why? If I use `multiplyBy` it will return `BigNumber` and cannot use `ceil` after that.

snowypowers: doing multiplication in js introduces inaccuracies. There is `times` or `mul`. have you tried that?

MorganDream: OK

packages/neon-api/src/transaction/validator.ts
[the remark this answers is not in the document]
snowypowers: in my experience in neo2, fault can happen because the contract method requires a signature check which an invokescript is unable to provide. However, the script is valid.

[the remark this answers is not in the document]
MorganDream: So shall we use `invokeScript` to validate script? Considering it may return `FAULT` in a valid script.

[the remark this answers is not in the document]
snowypowers: If it fails, we probably set a warning stating that it produced a fault. Autofix should also fail since we cannot accurately estimate the cost if it fails. We could just give the GAS used up to the failure as a baseline in the message.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * validate systemFee
+   * @param autoFix will automatically fix transaction if specified as true
+   */
+  public async validateSystemFee(autoFix = false): Promise<ValidationResult> {
+    const validateIntentsResult = await this.validateScript();
+    if (!validateIntentsResult.valid) {
+      return validateIntentsResult;
```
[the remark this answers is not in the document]
MorganDream: I removed validating script here.

packages/neon-api/__integration__/transaction/validator.ts
```diff
@@ -0,0 +1,388 @@
...
+        operation: "name",
+        args: []
+      });
+      const transaction = new TransactionBuilder({
+        script,
+        systemFee: 0
+      }).build();
+      const validator = new TransactionValidator(rpcClient, transaction);
+      const result = await validator.validateSystemFee();
+      expect(result.valid).toBeFalsy();
```
snowypowers: should there be an expectation of a suggestion?

packages/neon-api/__integration__/transaction/validator.ts
```diff
@@ -0,0 +1,388 @@
...
+        operation: "name",
+        args: []
+      });
+      const transaction = new TransactionBuilder({
+        script,
+        systemFee: 10.1
+      }).build();
+      const validator = new TransactionValidator(rpcClient, transaction);
+      const result = await validator.validateSystemFee();
+      expect(result.valid).toBeFalsy();
```
snowypowers: Similarly, there should be a suggestion here?

packages/neon-api/__integration__/transaction/validator.ts
```diff
@@ -0,0 +1,388 @@
...
+    expect(result.valid).toBeTruthy();
+  });
+
+  test("invalid", async () => {
+    const transaction = new TransactionBuilder({
+      validUntilBlock: 1
+    }).build();
+    const validator = new TransactionValidator(rpcClient, transaction);
+    const result = await validator.validateValidUntilBlock();
+    expect(result.valid).toBeFalsy();
```
snowypowers: suggestion?

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,344 @@
...
+    if (state === "FAULT") {
+      return {
+        valid: false,
+        suggestions: {
+          systemFee: {
+            ...{
+              fixed: false,
+              message: "Encountered FAULT when validating script."
+            },
+            ...result.suggestions?.systemFee
```
snowypowers: You might want to change this such that the suggestion only appears if the current attached fee is lower than what was consumed before it faulted.

packages/neon-api/__integration__/transaction/validator.ts
MorganDream: The suggested value is dynamically changing as it's related to block height.

packages/neon-api/__integration__/transaction/validator.ts
MorganDream: OK

packages/neon-api/src/transaction/validator.ts
MorganDream: Message changed

packages/neon-api/__integration__/transaction/validator.ts
snowypowers: You can check for the existence of the suggestion and it be bigger than the current number.

packages/neon-api/src/transaction/validator.ts
```diff
@@ -0,0 +1,345 @@
...
+      return {
+        valid: false,
+        result: {
+          systemFee: {
+            ...{
+              fixed: false,
+              message:
+                "Cannot get precise systemFee as script execution on node reports FAULT."
+            },
+            ...validation.result?.systemFee
```
snowypowers: I thought we discussed and this suggestion should only appear when the suggestion is smaller than the currently attached fee.

hal0x2328:
500 points MorganDream

lllwvlvwlll:

PR #491, opened 2019-11-25
feat(core): NeonObject

snowypowers:
This introduces the NeonObject interface and adds them to the original few objects that implements them.

A breaking change was introduced to Network and Protocol as they exported the configuration style object instead. Now, the export method will work as normal to export it as a JSON object and a new method toConfiguration is added that will export it as PascalCased JSON which was the old behavior for export.


hal0x2328:
400 points snowypowers

lllwvlvwlll:

PR #501, opened 2019-12-23
[fix] networkFee calculation is not right in validateNetworkFee.

MorganDream:
- Bug 
Networkfee calculation is based on the witnesses(signatures) of the transaction.
However the signatures are invalid before networkFee is set, so validation(calculation) of networkFee should be done before signing the transaction.
- Fix
validateNetworkFee API needs param `signers` - accounts that will sign this transaction.

### DIFF
--- a/packages/neon-api/__integration__/transaction/validator.ts
+++ b/packages/neon-api/__integration__/transaction/validator.ts
@@ -3,23 +3,18 @@ import {
   TransactionValidator,
   ValidationAttributes
 } from "../../src/transaction";
-import { sc, tx, u, rpc, CONST } from "@cityofzion/neon-core";
+import { sc, tx, u, rpc, CONST, wallet } from "@cityofzion/neon-core";
 
 const TESTNET_URL = "http://seed5t.neo.org:20332";
 const rpcClient = new rpc.RPCClient(TESTNET_URL);
 
-const sig = {
-  invocationScript:
-    "40f52d1206315dfac64c14ec2dfef1edd62f4460487c23be6bbbbf9080973784ca7dbfe4dfcf4b6b82f2921b968e0d693020b76be0b20171ac56e7da50ab1c4b06",
-  verificationScript:
-    "21031d8e1630ce640966967bc6d95223d21f44304133003140c3b52004dc981349c968747476aa"
-};
-const multiSig = {
-  invocationScript:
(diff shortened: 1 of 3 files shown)


MorganDream:
@snowypowers 
To build a transaction, you need only the scripthash/address info of `signers`(sender and cosigners). Verificationscript in witnesses is `account.contract.script`, which should be unnecessary if just considering constructing a transaction.

`TransactionBuilder` is now a independent class, if we adding verificationscript in witnesses, there is possibility that user may directly use it, what if user wants to sign transaction when there is such dummy witness? And the transaction to validate may be constructed directly by `Transaction` class.

So I think to keep an isolated design, current solution is a better way.

snowypowers:
Since validation of network fee will require the use of the verification script it's just a matter of whether are we gathering it at the build step first or at the validate step. 

I'm not clear why are you mentioning the builder being an independent class. My suggestion doesn't not change that fact.

You can sign a transaction with dummy witnesses since signatures sign the transaction hex string without the witness data.

My suggestion will still allow the transaction to be constructed correctly, it just appends more data to the object earlier than later. And if we are gathering the data for the validate step, why not use it at the build step?

MorganDream:
> Since validation of network fee will require the use of the verification script it's just a matter of whether are we gathering it at the build step first or at the validate step.
It's not required to gather verification script at build step. Build and validate steps are independent, so we should not gather something in build step which is not necessary in build step, just for another step- validate.

snowypowers:
It is possible that we populate the network fee in the builder too if it is fully offline. This looks like a job for the builder than the validator.

Having the information appended within the transaction makes it easy for each user to independently validate the transaction before signing. I will only have to transmit the transaction to everyone instead of the transaction and the list of accounts.

The data is going to be present in the transaction anyway upon signing so there is no extra overhead in adding this data.

Even as separate steps, the build and validate steps are going to be done closely together. I do not see an issue in gathering the accounts in the build step. It is also likely that the accounts can be provided as easily at either steps.

MorganDream:
> Even as separate steps, the build and validate steps are going to be done closely together.
One of the advantages to make processes independent is that it's flexible to arbitrarily combine the steps. Now `TransactionValidator` accepts 2 params: rpcClient and transaction, it indicates that I can use this class if I have a transaction and a network client. The transaction may be constructed by the basic `Transaction` constructor or even `getTransaction` rpc method, in which case the accounts info have not been gathered.

If we tell user validator must be used after builder, then why do we separate them?

snowypowers:
The separation is mainly for separation of concerns and flexibility of usage.

PR #506, opened 2020-02-12
fix(sc): Fix emitAppCall with only scriptHash

### LINKED ISSUE #502: Calling emitAppCall without args parameter adds an unnecessary PUSH0 to the script
If emitAppCall is called without the args parameter being specified, the script that is generated is prefixed with an unnecessary PUSH0 (0x00) byte. Internally the neon-js code treats the 'undefined' parameter as if it were a boolean false, so pushes a 0 to the stack.  This results in an unexpected 0 being left on the stack after the script is invoked.

Example:

import * as neon from '@cityofzion/neon-js';
// ...
const contractHash = '...'; // replace ... with some valid contract hash
const sb = neon.default.create.scriptBuilder();
const script = sb.emitAppCall(contractHash).str; // script is prefixed with an unexpected 00

snowypowers:
fixes #502

Originally, the call to emitAppCall without any arguments will produce a script that will push a `00` at the start of the script. This can be seen as an error if the contract did not consume the parameter, leaving it in the stack.

By adding a check for undefined, we can prevent the `00` from appearing when no arguments are provided. This breaks existing behaviour for all scripts that rely on no arguments filling in an empty parameter. For example, standard NEP5 contracts will consume 2 variables off the stack. This fix will cause the old code to emit only a single variable and result in a fault when running the script.

This can be fixed by explicitly providing an empty array for args (which is what should happen anyway) which will prompt the ScriptBuilder to inject the appropriate variable onto the stack.

Fundamentally, this is a design fault in exposing the emitAppCall in such a fashion. The current defaults makes it difficult to adapt the method to be flexible enough for contracts that do not follow the convention of `Main(string, any[])`.

### DIFF
--- a/packages/neon-core/__tests__/sc/ScriptBuilder.ts
+++ b/packages/neon-core/__tests__/sc/ScriptBuilder.ts
@@ -116,6 +116,13 @@ describe("emitAppCall", () => {
         useTailCall: false
       },
       "143775292229eccdf904f16fff8e83e7cffdc0f0ce51c10962616c616e63654f666711c4d1f4fba619f2628870d36e3a9773e874705b"
+    ],
+    [
+      "emitAppCall with only scriptHash",
+      {
+        scriptHash: "5b7074e873973a6ed3708862f219a6fbf4d1c411"
+      },
+      "6711c4d1f4fba619f2628870d36e3a9773e874705b"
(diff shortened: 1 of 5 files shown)


djnicholson:
👍

PR #508, opened 2020-02-29
fix(chore):Fixes all major security vulnerabilities with npm deps

comountainclimber:
Using `resolutions` in `package.json` I have manually addressed numerous npm dependency security issues. To test, checkout this branch and run `yarn audit` in either the website or main directories.

Screen Shot 2020-02-28 at 6 15 49 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,5 +45,12 @@
   "private": true,
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "resolutions": {
+    "handlebars": "^4.3.0",
+    "set-value": "^2.0.1",
+    "mixin-deep": "^1.3.2",
+    "lodash": "^4.17.12",
+    "https-proxy-agent": "^2.2.3"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -47,10 +47,10 @@
     "packages/*"
   ],
   "resolutions": {
-    "handlebars": "^4.3.0",
-    "set-value": "^2.0.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,5 +45,12 @@
   "private": true,
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "resolutions": {
+    "handlebars": "^4.3.0",
```
snowypowers: since we are resolving these specifically, should we pin them to an exact version? What is the plan going forward to remove the pins? Do I have to remove each pin and run yarn audit?

comountainclimber: We could pin them to an exact version this basically is saying that sub dependencies need to install at least the version or higher. If we pin them we will lose out on any updates to versions dependencies rely on so might not be the best choice.

>  Do I have to remove each pin and run yarn audit?

You dont have to do anything except run `yarn && yarn audit`

snowypowers: so we pin the dependencies to the exact version to have somewhat reproducible builds. I would prefer that we do the same for the resolutions. Ideally, we want to remove these resolutions when the primary dependency updates. I think yarn audit will always report ok when I run it with the resolutions.

comountainclimber: > I would prefer that we do the same for the resolutions. 

This is not really the right way to do this... We are losing out on all future dependency updates by pinning exact resolutions but I will go ahead and do it anyway

Issue #511, opened 2020-03-12
[feature] Add support for event notifications


corollari:
## Motivation
When building interfaces for smart-contract-based applications it's pretty common to want it to react to changes in the contract state in order to make the UI accurately represent the current state. Examples of this could be an exchange that wants to remove an outstanding buy order from the interface when such order is taken or withdrawn, or a game that would display the moves of an opponent the moment they happen.

Right now this can only be achieved through constant querying of the RPC nodes or by creating a custom solution, but something that would serve this need much better would be a system that would allow a developer to subscribe to a smart contract and be notified of event triggers on it, essentially a publish-subscribe system.

As a data point, on Ethereum, a system like this is integrated into all the major libraries and used by almost all dApps. Note that I'm not advocating for this library to become a copy of web3, but pointing out that in a similar network this kind of feature has seen a lot of usage.

## Feature
Integrate a publish-subscribe framework that enables developers to subscribe to a smart contract in order to receive events triggered on it, an example of such api could be:
```
import { api } from "@cityofzion/neon-js";

api.pubsub.subscribe(address, (event)=>doSomething(event));
```

where the second argument is a callback function that would be called every time an event in that contract is triggered.

This is just an idea to illustrate the concept tho, the actual implementation can follow other designs.

## Infrastructure
Unless we turn neon-js clients into nodes capable of interacting with the p2p network and executing scripts on the NeoVM, integrating this feature into neon-js will require some extra infrastructure that collects events being triggered inside smart contracts and broadcasts them to subscribed clients, essentially the publisher of pubsub architecture.

I've written, maintain and run [a server that could serve that purpose](https://github.com/corollari/neo-PubSub). In it's current form it broadcasts events triggered on all neo smart contracts and messages received through the p2p network. Furthermore, the architecture it uses consists of a single neo node that collects events and messages and broadcasts these to a dynamic set of servers running on heroku (this set can scale with demand) which maintain websocket connections with all the subscribers and rebroadcast everything to them, resulting in a highly scalable system, which should be able to support heavy loads.

The system could be made more scalable by using AWS IoT to relay the messages, but I don't think it will reach a load where that is necessary, and if we do then we can migrate it over there.

- Source code: https://github.com/corollari/neo-PubSub/
- Showcase page: https://corollari.github.io/neo-PubSub/

## Extra
Something that might also be interesting to implement is some mechanism that allows a developer to know when a transaction has been finalised, that is, has been included into a block.

[...]

PR #514, opened 2020-03-20
feat(api): Add notifications API

corollari:
This PR integrates the notifications API discussed in #511

Regarding the Notifications class, I haven't implemented an api like `api.neoscan.getBalance("http://www.neoscan-testnet.io/test_net/v1/", address)` because the `notifications` api is stateful (needs to keep track of the websockets connections to be able to close them later) while `neoscan` is largely not.


packages/neon-api/src/provider/notifications/class.ts
```diff
@@ -0,0 +1,48 @@
...
+      this.url = settings.networks[url].extra.notifications;
+    } else {
+      this.url = url;
+    }
+	this.websocketSubscriptions = new Map<string, WebSocket[]>();
+    log.info(`Created Notifications Provider: ${this.url}`);
+  }
+
+  public subscribe(contract: string, callback: (message: NotificationMessage) => void) {
+	  const ws = new WebSocket(this.url + "?contract=" + contract);
```
snowypowers: new websocket for every callback? We can do better than this. For subscriptions to the same contract, I expect that we reuse the old socket.

packages/neon-api/src/provider/notifications/class.ts
```diff
@@ -0,0 +1,48 @@
...
+  private url: string;
+
+  public get name(): string {
+    return `Notifications[${this.url}]`;
+  }
+
+  private websocketSubscriptions: Map<string, WebSocket[]>;
+
+  public constructor(url: string) {
+    if (settings.networks[url] && settings.networks[url].extra.notifications) {
```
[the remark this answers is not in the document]
corollari: This was taken from the code of other providers, we'd have to change that other code also in order to maintain consistency

packages/neon-api/src/provider/notifications/class.ts
corollari: Agree, working on it.

packages/neon-api/src/provider/notifications/class.ts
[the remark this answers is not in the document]
snowypowers: no worries. Focus on this file for now. We can migrate the rest in another PR. This is a new feature that came with Typescript 3.7.

corollari:
> I would avoid the use of the word Provider here as there is a specific interface for it. Reusing the name would be confusing. Consider not calling it a Provider in the docs and moving it under a different folder.

Makes sense.

> Any reason why we are not adding support for the rest of the endpoints or is that coming too?

Looking at the evolution and usage that this kind of system has had on ethereum it seems that users mostly just have two use-cases:
- Listen for events in their own smart contract
- Activate a callback when a transaction is consolidated in a block

Because of this I think that the other endpoints should be integrated in different ways (directly into the API that broadcasts transactions for example) but I believe that this other part is more controversial and wanted to get the first PR simple, that's why for now I'm only integrating a single endpoint.

> Please try to separate the code and documentation. Docs can come later when we commit the code. If there are big changes, its gonna be a headache trying to reread the docs every time something changes.

Okay.

packages/neon-api/src/provider/notifications/class.ts
```diff
@@ -0,0 +1,48 @@
...
+      this.url = settings.networks[url].extra.notifications;
+    } else {
+      this.url = url;
+    }
+	this.websocketSubscriptions = new Map<string, WebSocket[]>();
+    log.info(`Created Notifications Provider: ${this.url}`);
+  }
+
+  public subscribe(contract: string, callback: (message: NotificationMessage) => void) {
```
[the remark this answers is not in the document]
corollari: Personally, when designing APIs I believe the most important thing is to adhere to standards that developers are already familiar with, as that makes the API much more easier to use in my experience. With that in mind, I came up with three different models when designing this api:
1. The model on this PR: `subscribe(contract, callback), unsubscribe(contract)`
2. The standard subscription model used in DOM APIs: `subscribe(contract, callback), unsubscribe(contract, callback)`
3. The model you proposed: `a = subscribe(contract, callback); a.unsubscribe();`

While the 3rd model is the one that I find the most elegant, I believe that the other two would be more intuitive to the developers using neon-js, because:
- 2nd model: A lot of them are probably familiar with the DOM api, given that they are using a js library
- 1st model: This is the model used by [the dAPI](https://neologin.io/api/#event-methods). I didn't design it and I believe it's not the best model but in the future I expect that a lot of the devs using neon-js on the front-end will also have to deal with the dAPI, so there's some value in having consistency there.
- 1st model: I searched "pubsub javascript" on google and two of the first results that came up were tutorials that taught how to implement a pubsub system based on the 1st model

Because of this, I chose to go with the 1st model, but I have no problem changing it if it makes more sense.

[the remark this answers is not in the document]
corollari: > Subscribe to all contract events using null

The reason why I didn't account for this use-case is because I believe almost nobody will use it like this, but the solution you proposed (using null) doesn't pollute the api so I'm down for adding that.

corollari:
I believe I've now addressed all the concerns except those related to api design.

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  private websocketSubscriptions: Map<string | null, WebSocket[]>;
+
+  public constructor(url: string) {
+    this.url = settings.networks[url]?.extra?.notifications ?? url;
+    this.websocketSubscriptions = new Map<string | null, WebSocket[]>();
+    log.info(`Created Notifications Provider: ${this.url}`);
+  }
+
+  public subscribe(
```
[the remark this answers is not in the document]
corollari: Regarding that, I'm not really sure when to write a description for a method and when not to. Looking at the current source the only pattern I can see is that functions with lots of lines are generally annotated, is there any heuristic in place for that?

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,49 @@
...
+    contract: string | null,
+    callback: (message: NotificationMessage) => void
+  ) {
+    const ws = new WebSocket(
```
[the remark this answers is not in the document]
corollari: Yeah, given that I'll probably have to re-write everything when we change the api design, I was waiting for that.

corollari:
I see, what kind of object were you thinking to use as `reference` on the unsubscribe calls?

packages/neon-api/src/notifications/class.ts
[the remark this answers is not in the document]
snowypowers: generally I would annotate the public api. So it would be the constructor, subscribe and unsubscribe here. Unsubscribe might be pretty straight forward so no need but the constructor and subscribe would benefit from some description to indicate how to use. For example, indicating that you can subscribe to all by not passing an argument.

snowypowers:
I was thinking of just an empty object at the moment. You may consider adding maybe an identifier (maybe human readable). This paves the way for us to transform this into a class with methods if we ever need to go down the route where we want the unsubscribe/pause function on the reference.

corollari:
So something that would work like this:
```
let object = notificationsProvider.subscribe('0xabc')
notificationsProvider.unsubscribe(object)
```
?

packages/neon-api/src/notifications/class.ts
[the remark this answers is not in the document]
corollari: Makes sense, I thought that would be the case but when I looked at the code of the neoCli provider there was only a single function annotated, an internal one not directly exposed. But I can see that the functions there have more descriptive names so it would make sense not to annotate them.

snowypowers:
yeah

corollari:
I'm currently busy with exams so I'll put this PR on hiatus and come back to it in a few weeks.

corollari:
I've changed the API to work in the way you initially proposed:
```
let subscription = notifications.subscribe(hash, console.log)
subscription.unsubscribe()
```

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+   * Subscribe to event notifications of a specific contract
+   * @param contract - Hash of the contract (null for all contracts) to subscribe to
+   * @param callback - Function to call when a notification is received.
+   * @return Subscription object that can be used to cancel the subscription
+   */
+  public subscribe(
+    contract: string | null,
+    callback: CallbackFunction
+  ): Subscription {
+    if (this.subscriptions.has(contract)) {
```
snowypowers, line 41: Optional, a cleaner flow would be to first pull out the `ContractSubscriptions` object first and just do a push. 

```
const contractSubscriptions = this.subscriptions.get(contract) ?? createWebsocketForContract(contract);
contractSubscriptions.callbacks.push(callback);
```

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+      } else {
+        this.unsubscribeContract(contract);
+      }
+    };
+    return new Subscription(contract, unsubscribe);
+  }
+
+  /**
+   * Unsubscribe all callbacks from a specific contract
+   * @param contract - Hash of the contract (null for all contracts) to unsubscribe callbacks from
```
snowypowers, line 79: we will have to make it clear that the null string is a special subscription that listens to all contracts and `unsubscribeContract() !== unsubscribeAll()`.

packages/neon-api/src/notifications/subscription.ts
```diff
@@ -0,0 +1,32 @@
...
+    this.unsubscribeFunction = unsubscribeFunction;
+    this.alive = true;
+    log.info(`Created Subscription: ${this.contract}`);
+  }
+
+  /**
+   * Unsubscribe a specific function from a contract
+   */
+  public unsubscribe() {
+    if (!this.alive) {
```
snowypowers, line 24: this inner variable feels redundant when we are not throwing. The savings is pretty small for the cost of having a possible wrong state.

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    const unsubscribe = () => {
+      if (!this.subscriptions.has(contract)) {
+        // Needed because user might have called unsubscribeAll() before
+        return;
+      }
+      const subscriptions = this.subscriptions.get(contract)!;
+      const index = subscriptions.callbacks.indexOf(callback);
+      if (index === -1) {
+        // Could happen if unsubscribeAll() followed by subscribe() are called on the same contract
```
snowypowers, line 65: add a test for this edge case.

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+        }
+      };
+      this.subscriptions.set(contract, {
+        websocket: ws,
+        callbacks: [callback]
+      });
+    }
+    const unsubscribe = () => {
+      if (!this.subscriptions.has(contract)) {
+        // Needed because user might have called unsubscribeAll() before
```
snowypowers, line 59: testcase

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+        return;
+      }
+      const subscriptions = this.subscriptions.get(contract)!;
+      const index = subscriptions.callbacks.indexOf(callback);
+      if (index === -1) {
+        // Could happen if unsubscribeAll() followed by subscribe() are called on the same contract
+        return;
+      }
+      if (subscriptions.callbacks.length > 1) {
+        subscriptions.callbacks.splice(index, 1);
```
snowypowers, line 69: testcase

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+      } else {
+        this.unsubscribeContract(contract);
```
snowypowers, line 71: testcase

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+    return new Subscription(contract, unsubscribe);
+  }
+
+  /**
+   * Unsubscribe all callbacks from a specific contract
+   * @param contract - Hash of the contract (null for all contracts) to unsubscribe callbacks from
+   */
+  public unsubscribeContract(contract: string | null) {
+    if (!this.subscriptions.has(contract)) {
+      return; // Not throwing an exception in order to allow unlimited calling of this function
```
snowypowers, line 83: testcase

packages/neon-api/src/notifications/class.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    this.subscriptions.get(contract)!.websocket.close();
+    this.subscriptions.delete(contract);
+  }
+
+  /**
+   * Unsubscribe all callbacks from all contracts
+   */
+  public unsubscribeAll() {
```
snowypowers, line 92: testcase

packages/neon-api/__tests__/notifications/class.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(result.name).toMatch(expectedUrl);
+  });
+
+  test("Network name", () => {
+    const result = new notifications("UnitTestNet");
+    expect(result.name).toMatch(UnitTestNetUrl);
+  });
+});
+
+const contract = "0x314b5aac1cdd01d10661b00886197f2194c3c89b";
```
snowypowers, line 33: does it work with only 0x-prefixed strings? This detail should be noted.

packages/neon-api/src/notifications/subscription.ts
corollari: What alternative do you think would work best? Throwing exceptions?

snowypowers: I meant that you could just remove the `alive` check and just call `unsubscribeFunction` since it doesn't fail.

corollari: I see what you are saying, but that could result in unexpected behaviour in the following case:
```
let s1 = notifications.subscribe("0x...", console.log)
let s2 = notifications.subscribe("0x...", console.log)
s1.unsubscribe()
s1.unsubscribe() //s2 gets unsubscribed
```

With that said, this is a sort of an extreme edge case so we may want to ignore it.

snowypowers: ah thats interesting then the issue here is that we are storing the reference to the callback as an unique identifier. This solves the problem by ensuring each subscription can only at most remove one callback. A little roundabout but i guess it works.

corollari:
I'm currently reworking how the class synchronises the state of the subscriptions object with the individual subscriptions by storing callbacks along a unique identifier, therefore making it much easier to check whether the subscription has been removed before.

I'd thought about this previously but decided against it because I thought that looping over an array of callbacks (the most frequent action due to the fact that it's done every time a new event is received) would be faster compared to iterating over some other data structure such as the values of a map. Later, [your comment](https://github.com/CityOfZion/neon-js/pull/514#discussion_r407116381) made me consider that I might be wrong, so I ran a [performance benchmark](https://jsben.ch/leGlO) and, contrary to my expectations, it turns out that iterating over the values of a map only results in a slowdown of <2%, a performance hit that, in my opinion, is worth the simpler implementation, which results in less edge cases and removes the need to keep two components synchronized.

packages/neon-api/src/notifications/class.ts
corollari: Done in 50387ee49184e0db7caf59e2995ee40c5ad183c3

packages/neon-api/src/notifications/class.ts
corollari: Change the comments in ca90ea9508e29003659ea4e1e1a3395b4aa90b07 to try to make that clearer, thoughts?

packages/neon-api/src/notifications/subscription.ts
corollari: I've changed how that works to make reduce the edge cases and the complexity, see https://github.com/CityOfZion/neon-js/pull/514#issuecomment-626260820

corollari:
Just added a bunch of tests, let me know if I'm missing anything.

corollari:
With the addition of the last 2 tests, the coverage right now is sitting at 100%.

snowypowers:
finally done! Thanks for the contribution!

PR #530, opened 2020-05-14
fix(api): Add ws for notifications to work in Node

snowypowers:
- Install ws as a dependency
- Add integration smoke test
- Fix small bug where messages could still come in as close does not reject messages in buffer

### DIFF
--- a/packages/neon-api/__integration__/notifications/class.ts
+++ b/packages/neon-api/__integration__/notifications/class.ts
@@ -1,0 +1,25 @@
+import * as neonCore from "@cityofzion/neon-core";
+import apiPlugin from "../../src/index";
+
+const neonJs = apiPlugin(neonCore);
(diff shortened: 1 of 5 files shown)


corollari:
I see I've made a few mistakes on the typescript types and there's some code that could be simplified, but to me it seems that, functionally, the code remains the same. Could you point out which change fixed that bug? Or was that fixed through a change in dependencies? Also, could you share the reasoning behind adding `test.html`?

snowypowers:
[this change](https://github.com/CityOfZion/neon-js/pull/530/files#diff-4a3ec44543c13bc78a2f877138539f35R94) allows for the websocket to retrieve deleted callbacks as an empty array. The test failed once during teardown when the socket was closed and i got an error regarding callbacks being null. 

`test.html` was just a simple manual check for myself to see if I can use the package in a browser environment.

corollari:
I see, you're right.

PR #540, opened 2020-06-12
Feature/docs

lllwvlvwlll:
This enhancement adds documentation build and deploy to the new static documentation source as part of the CI build and deploy release process.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,6 +174,23 @@ jobs:
           name: Publish to npm
           command: yarn release:next --yes
 
+  publish-docs:
+    executor: env-node10
+    steps:
+      - checkout
+      - run:
+          name: install aws cli
+          command: |
+            curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+            unzip awscliv2.zip
+            sudo ./aws/install
+      - run:
+          name: Build and release docs
+          command: |
+            cd website
+            yarn install
+            yarn run release
+
 workflows:
   version: 2
   build_and_test:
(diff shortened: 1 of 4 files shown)


snowypowers:
hmm weird that circleci did not kick off any pipeline for this PR. I cant really confirm that the yaml parses correctly even if it looks correct.

I guess there should be also a change (in a separate PR) to update the master branch to also publish to s3?

.circleci/config.yml
```diff
@@ -174,6 +174,23 @@ jobs:
           name: Publish to npm
           command: yarn release:next --yes
 
+  publish-docs:
```
snowypowers, line 177: if we intend to continue publishing neo2 docs, we should name this `publish docs-next` or something to signal it.

.circleci/config.yml
```diff
@@ -211,14 +228,15 @@ workflows:
...
+          context: aws
           requires:
             - setup
           filters:
             branches:
-              only: master
+              only: next
             tags:
-              only: /v[0-9]+\.[0-9]+\.[0-9]+$/
+              only: /v[0-9]+\.[0-9]+\.[0-9]+-next.*/
```
snowypowers, line 239: Docs aren't necessary tied to version tags. We may wish to continually update docs without publishing a new version. Unfortunately, CircleCI doesnt have a good way to manually trigger workflows.

Optional: Consider triggering this workflow on every commit on the our release branches but have approval hold so we can dictate when to push a new version of docs.

.circleci/config.yml
```diff
@@ -174,6 +174,23 @@ jobs:
           name: Publish to npm
           command: yarn release:next --yes
 
+  publish-docs:
+    executor: env-node10
+    steps:
+      - checkout
+      - run:
+          name: install aws cli
```
snowypowers, line 182: Optional: we can be cool and try https://circleci.com/orbs/registry/orb/circleci/aws-s3.

.circleci/config.yml
snowypowers: I also strongly encourage this change so you can actually try this change out without publishing `-next.999`. I get 5 emails every publish. Thanks.

PR #565, opened 2020-08-18
feat(u): Rework Fixed8 to use bn.js

snowypowers:
This removes the dependence on `bignumber` package and consolidates the dependency to share it with elliptic. Should result in a smaller package size.


snowypowers:
nvm it doesnt do decimals well. its gonna be a pain to do this correctly.

PR #566, opened 2020-08-19
hotfix: neon-js/getRPCEndPoint failed

bokuns:


Tommo-L:
@snowypowers Could help merge this PR?  As it always return a empty array, and Qlink project want to use this sdk.

snowypowers:
The API client when receiving an empty list, throws an error signalling that no good endpoint can be found. This is working as intended. The fix you are proposing hides the fact and may return an invalid url which will cause breakage further down the flow.

From my perspective, the issue is that neoscan is broken and the fix should be either be:
1. Contact neoscan and get it fixed
2. Try-catch within your app and default to your own list of endpoints.

Option 2 seems to be the quickest fix for you. I would like to know why is that not preferable over you submitting this fix and waiting for a release cycle.

Tommo-L:
> Try-catch within your app and default to your own list of endpoints.

Could give an example?

snowypowers:
```
let rpcEndpoint = null;
try {
const neoscan = new Neon.api.neoscan.instance("TestNet");
rpcEndpoint = await neoscan.getRPCEndpoint();
} catch {
rpcEndpoint = defaultUrl;
}

//continue using rpcEndpoint
```

lllwvlvwlll:
I agree with @snowypowers that we want an explicit failure as currently implemented.  @Tommo-L will his example work for you?

Tommo-L:
Thanks, it's fixed. Qlink uses http://cityofzion.io/neon-js/docs/en/api/core/rpc.html

Issue #579, opened 2020-09-25
extend OpToken for full disassembly support


ixje:
https://github.com/CityOfZion/neon-js/pull/577 added partial VM script disassembling. I think the functionality could be useful beyond the original reason for creation (multisig parsing). As it stands no translation to the actual syscall names is done, `JMP_x`  and a handful or other instructions will fail to parse. If any of the missing instructions come before the currently supported PUSHINT/PUSHDATA then the remainder of the script will be disassembled incorrectly.

For reference; here is a full NEO3 VM script disassembler https://github.com/CityOfZion/neo3-preview/pull/77

snowypowers:
At the current state, it should be able to breakdown any script. As far as I know, only the PUSH* codes have additional params that require work, the rest are simple opcodes that dont touch the script at all. If there are any, feel free to add the logic. 

Currently:
```
Neon.sc.OpToken.fromScript(Neon.u.HexString.fromBase64("DAUA6HZIFwwUZN9OvpIzTR/H5kvx0eM5QtmQXlEMFM5hb390YX4PxLgFWDryYCojjfY/E8AMCHRyYW5zZmVyDBQwP75r/oFnA5XLgzPAxIoULlM+hkFifVtS").toBigEndian()).map(t => t.toInstruction())
[
  'PUSHDATA1 0500e8764817',
  'PUSHDATA1 1464df4ebe92334d1fc7e64bf1d1e33942d9905e51',
  'PUSHDATA1 14ce616f7f74617e0fc4b805583af2602a238df63f',
  'PUSH3',
  'PACK',
  'PUSHDATA1 087472616e73666572',
  'PUSHDATA1 14303fbe6bfe81670395cb8333c0c48a142e533e86',
  'SYSCALL 627d5b52'
]
```

Notable differences:
- I retained the length byte as part of the params. Probably up for debate whether this is good or bad but it is trivial to change the implementation. A benefit of the current implementation is that reassembling the script is relatively simple.
- I did not translate the syscall to an interop call name.
- No pretty formatting.

One way we can go is to retain how I chose to store the info (`params` field) but for `toInstruction` to remove the length byte and do the interop name translation. AKA change `toInstruction` to produce output similar to your disassembler.

ixje:
* > At the current state, it should be able to breakdown any script.

[...]

PR #585, opened 2020-09-28
feat(sc): Add disassembly support for all OpTokens

snowypowers:
- Add OpCodeAnnotations containing data from the annotations in C#.
- Add more tests around new OpCodes.
- Rename toInstruction to prettyPrint.
- Add padding for prettyPrint.
- Remove length byte from being included in params.

#579


packages/neon-core/src/sc/OpToken.ts
```diff
@@ -47,9 +53,7 @@ export class OpToken {
     if (opToken.code >= 0 && opToken.code <= 5) {
       //PUSHINT*
       // We dont verify the length of the params. Its screwed if its wrong
-      return opToken.params
-        ? HexString.fromHex(opToken.params, true).toNumber()
-        : 0;
+      return opToken.params ? parseInt(reverseHex(opToken.params), 16) : 0;
```
ixje, line 56: Give me a second , I'm about to look at what the effect of this is

snowypowers: This is just a refactor that removes the need to use HexString which is a class vs using direct methods. Should be equivalent. Just using a lower level import.

ixje: Sorry took a bit longer because I had modification in my C# environment that was giving me weird output.

Following the C# [unit tests](https://github.com/neo-project/neo-vm/blob/master/tests/neo-vm.Tests/Tests/OpCodes/Push/PUSHINT8_to_PUSHINT256.json#L47), the following should output `-2`, instead it gives `65534`

```javascript
const token = Neon.sc.OpToken.fromScript("01feff");
let value = Neon.sc.OpToken.parseInt(token[0]);
console.log(value);
```

Values you use to test
PUSHINT8,  00ff, expected: -1
PUSHINT16, 01feff, expected: -2
PUSHINT32, 03fdffffff, expected: -3
PUSHINT64, 04fcffffffffffffff, -4
PUSHINT128, 05fbffffffffffffffffffffffffffffff, -5
PUSHINT256, 06faffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, -6

ixje: fwiw; this `Buffer.from('feff', 'hex').readIntLE(0,2)` produces the correct number

snowypowers: looks like i have to work on getting the big integer representation in first for this to be implemented well.

ixje: can you share some info on what `getting the big integer representation in first` this means? Do you need to find a biginteger library or? If so, we might consider wrapping this [C++ port](https://github.com/CityOfZion/BigIntegerCpp) of the C# Biginteger class. It was specifically made to produce the identical output.

snowypowers: I was looking at https://github.com/NeoResearch/csbiginteger-js.

The C++ port is not suitable as that cannot be packaged up for web. Alternatively is just writing some small wrapping functions for reading hexstrings as two complements.

ixje: I don't want to down talk their work, but I've historically tried to use their biginteger library in python and it produced wrong output in certain cases. What I remember is one option just wrapped a different C++ library, the other had some interface to talk to C# but now you need both .NET and the bindings. For these reasons I did a full native port with 0 dependencies. I'm positive it can be packaged up for web because the javascript's core is C++.

That aside if you think the existing code just needs a couple of small functions, then it's probably not worth the effort to create bindings at this moment.

snowypowers: updated to use BigInteger.

ixje: will check soon

ixje: It does not seem to like 
```javascript
const token = Neon.sc.OpToken.fromScript("06faffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
```
throwing
```
/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:30659
            const paramsExtracter = annotation.operandSize
                                               ^

TypeError: Cannot read property 'operandSize' of undefined
```

while the following does work as expected
```javascript
const token = Neon.sc.OpToken.fromScript("05fbffffffffffffffffffffffffffffff");
```

ixje: The rest looks good to me.

snowypowers: I believe PUSHINT256 is `05`. The test case is here: https://github.com/CityOfZion/neon-js/pull/585/files#diff-2a331f7ce64869df2deb51b6707975e5R117

ixje: You're right, my bad

ixje (approved):

ixje:
🎉

PR #588, opened 2020-10-04
fix(rpc): Guard against non-http urls

### LINKED ISSUE #587: [enhancement] validate RPCClient constructor arguments
I called the RPCClient constructor with as parameter "127.0.0.1:10332" and got a nasty message somewhere deep down the call stack
```
TypeError: Cannot read property 'replace' of null
    at dispatchHttpRequest (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:192:74)
    at new Promise (<anonymous>)
    at httpAdapter (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:129:10)
    at dispatchRequest (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:1057:10)
    at async sendQuery (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:29136:22)
    at async RPCClient.execute (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:29159:29)
    at async RPCClient.invokeScript (/Users/erik/Documents/code/neon-js/packages/neon-js/dist/index.js:29034:20)
``` 

I'd like to suggest an enhancement where the "net" parameter of the constructor is validate for correct formatting and return a friendlier error message if used incorretly.

snowypowers:
Guard against scheme-less urls as axios does not handle them well on server side.

A simple string match will do for us. We currently only operate on http/https protocol so we can just ensure all urls start with it.

Fixes #587

### DIFF
--- a/packages/neon-core/src/rpc/clients/RpcDispatcher.ts
+++ b/packages/neon-core/src/rpc/clients/RpcDispatcher.ts
@@ -49,6 +49,9 @@ export class RpcDispatcher {
   public url: string;
 
   public constructor(url: string) {
+    if (!url.startsWith("http")) {
+      throw new Error("Please provide an url that starts with http or https");
+    }
     this.url = url;
   }
 
--- a/packages/neon-core/src/rpc/clients/RpcDispatcher.ts
+++ b/packages/neon-core/src/rpc/clients/RpcDispatcher.ts
@@ -49,8 +49,10 @@ export class RpcDispatcher {
   public url: string;
 
   public constructor(url: string) {
-    if (!url.startsWith("http")) {
-      throw new Error("Please provide an url that starts with http or https");
+    if (!url.startsWith("http://") && !url.startsWith("https://")) {
+      throw new Error(
+        "Please provide an url that starts with http:// or https://"
+      );
     }
     this.url = url;
   }
--- a/helpers/urls.ts
+++ b/helpers/urls.ts
@@ -60,10 +60,10 @@ export async function getUrl(net: string): Promise<string> {
   let previousBlockCount = 0;
   for (let i = 0; i < slicedUrls.length; i++) {
     try {
-      const res = (await rpc.Query.getBlockCount().execute(slicedUrls[i])) as {
-        result: number;
-      };
-      const currentBlockCount = res.result;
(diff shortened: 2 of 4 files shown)


packages/neon-core/src/rpc/clients/RpcDispatcher.ts
```diff
@@ -49,6 +49,9 @@ export class RpcDispatcher {
   public url: string;
 
   public constructor(url: string) {
+    if (!url.startsWith("http")) {
+      throw new Error("Please provide an url that starts with http or https");
```
ixje, line 53: Overall PR looks good to me 👍

Just an open question here to understand a bit of your design philosophy; any specific reason you've not chosen to be slightly more strict here? As in something like
```javascript
if (!url.startsWith("http://") && !url.startsWith("https://")) {
```
I can imagine a `http:/` (single `/`) is easy to make.

snowypowers: The fix was merely to guard against people forgetting to append the `http` scheme for the url. Solving for confusion (people are used to `localhost` urls working without the http scheme attached) vs typos (`http:/` is a clear visible typo).

ixje: I'm getting strange behaviour though if I make the typo
```
const rpc_client = new Neon.rpc.RPCClient("http:/127.0.0.1:10332");
rpc_client.getBlockCount().catch((reason) => {
  console.log(reason);
});
```
Note how the port has changed to `:80` just because I forgot a `/`. That would throw me off wondering where the address came from.
```
Error: connect ECONNREFUSED 127.0.0.1:80
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1145:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 80,
  config: {
    url: 'http:/127.0.0.1:10332',
    method: 'post',
```

ixje: actually it is worse if you happen to not use `127.0.0.1`, try something like `http:/1.2.3.4`.

ixje (approved):
Thanks for the consideration

Issue #581, opened 2020-09-25
System and Network Fee Calculations


lllwvlvwlll:
from https://github.com/CityOfZion/neon-js/issues/571#issuecomment-698908831

I've had a look at the following item
> [High Priority] system and network fee calculations

The gas consumption for system fee can be obtained via RPC using [invokescript](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokescript.html). The network fee is where it gets interesting. 

Network fee = transaction base size + sum(signer's scripts) * fee per byte. ([ref](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L369-L432))

A signer script can be 1 of 3 things
1) a single signature redeem script
2) a multi signature redeem script
3) an actual smart contract with a verification function

The fee of the first 2 can be statically determined ([ref](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L410-L426)). The fee of the smart contract (3) requires [running the VM](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L397-L408) (similar to what we need todo for system fee). The problem here is that it needs to run the VM using the `Verification` trigger [type](https://github.com/neo-project/neo/blob/588cd983250f97659c6f91d7cd76ca7ff7e82990/src/neo/SmartContract/TriggerType.cs#L9-L16), whereas all RPC calls only allow running with the `Application` trigger type. 

Assuming I didn't miss anything the only short term feasible solution I see is having an RPC call that allows running the VM with the Verification trigger type. This can either be modifying the `invokescript` call to allow specifying a trigger type or a whole new RPC call. Anyone sees a different solution? @lllwvlvwlll @snowypowers 
If not, then I'll start the discussion on the NEO project repository.

### Task break down
As for supporting what is currently possible, I see the following tasks
- [ ] fix transaction size calculation.

   e.g. the [Signer](https://github.com/CityOfZion/neon-js/blob/1d9f61d7c1db4f9c202f0120b53ac38bf73fd51b/packages/neon-core/src/tx/components/Signer.ts) class currently returns a wrong size if `allowedContracts` or `allowedGroups`  is specified. 

[...]

Issue #571, opened 2020-09-03
Neo3 Discussion


lllwvlvwlll:
This issue can be used to discuss and break down outstanding features for the initial Neo3 release of neon.js

Current needs are as follows:
* [High Priority] api package maturity (this can be further defined using an audit of the neon wallet)
* [High Priority] system and network fee calculations
* [High Priority] documentation maturity
* [Medium Priority] tsdoc
* [Medium Priority] contract manifest-based field validation
* [Medium Priority] test expansion
* [Low Priority] notifications (this may be a feature we could add support for in a later release by adding dora api support)

@snowypowers @comountainclimber let me know if I missed anything or there is any discrepancy on priority.

ixje:
I've had a look at the following item
> [High Priority] system and network fee calculations

The gas consumption for system fee can be obtained via RPC using [invokescript](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokescript.html). The network fee is where it gets interesting. 

Network fee = transaction base size + sum(signer's scripts) * fee per byte. ([ref](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L369-L432))

A signer script can be 1 of 3 things
1) a single signature redeem script
2) a multi signature redeem script
3) an actual smart contract with a verification function

The fee of the first 2 can be statically determined ([ref](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L410-L426)). The fee of the smart contract (3) requires [running the VM](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L397-L408) (similar to what we need todo for system fee). The problem here is that it needs to run the VM using the `Verification` trigger [type](https://github.com/neo-project/neo/blob/588cd983250f97659c6f91d7cd76ca7ff7e82990/src/neo/SmartContract/TriggerType.cs#L9-L16), whereas all RPC calls only allow running with the `Application` trigger type. 

Assuming I didn't miss anything the only short term feasible solution I see is having an RPC call that allows running the VM with the Verification trigger type. This can either be modifying the `invokescript` call to allow specifying a trigger type or a whole new RPC call. Anyone sees a different solution? @lllwvlvwlll @snowypowers 
[...]

PR #603, opened 2020-10-08
feat(u): Add getVarSize helper function

ixje:
This function is necessary to help resolve the network fee calculation issue #581. The previously [suggested alternative](https://github.com/CityOfZion/neon-js/issues/571#issuecomment-699088064) of first serializing the object and taking the length cannot be used for the mentioned use case because part of the gas calculation is as follows:
```C#
int size = Transaction.HeaderSize + tx.Signers.GetVarSize() + tx.Attributes.GetVarSize() + tx.Script.GetVarSize() + IO.Helper.GetVarSize(hashes.Length);
```
Note how witnesses are not part of the size calculation where they would be if we called `size()` on the Transaction object. The exact reason of why the calculation is like it is can be found in the 2nd paragraph [here](https://github.com/neo-project/neo/issues/840#issue-457034133)

### DIFF
--- a/packages/neon-core/src/u/misc.ts
+++ b/packages/neon-core/src/u/misc.ts
@@ -1,4 +1,5 @@
 import { ab2hexstring } from "./convert";
+import { HexString } from "./HexString";
 
 const hexRegex = /^([0-9A-Fa-f]{2})*$/;
 
@@ -77,4 +78,40 @@ export function reverseHex(hex: string): string {
     out += hex.substr(i, 2);
   }
   return out;
+}
+
+/**
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
+export function getVarSize(value: any): number {
+  if (typeof value === "object" && value instanceof HexString) {
+    const size = value.byteLength;
+    return getVarSize(size) + size;
+  } else if (typeof value === "number") {
+    if (value < 0xfd) return 1;
+    else if (value <= 0xffff) return 3;
+    else return 5;
+  } else if (Array.isArray(value)) {
+    const array_length = value.length;
+    let size = 0;
+    if (array_length > 0) {
+      const tmp = value.map((item) => {
+        // in lack of an ISerializable interface, we check it like this
+        if (
+          typeof item.size === "number" &&
+          typeof item.serialize === "function"
+        ) {
+          return item.size;
+        } else if (typeof item === "object" && item instanceof HexString) {
+          return item.byteLength;
+        } else {
+          throw new Error("Unsupported value type in array: " + typeof value);
+        }
+      });
+      size = tmp.reduce((a, b) => a + b);
+    }
+    return getVarSize(array_length) + size;
+  } else {
+    throw new Error("Unsupported value type: " + typeof value);
+  }
 }
--- a/packages/neon-core/src/u/misc.ts
+++ b/packages/neon-core/src/u/misc.ts
@@ -83,7 +83,7 @@ export function reverseHex(hex: string): string {
 /**
  * Calculates the byte size of any supported input following NEO's variable int format.
  */
-export function getVarSize(value: any): number {
+export function getSerializedSize(value: any): number {
   console.log(typeof value);
   switch (typeof value) {
     case "number": {
@@ -94,22 +94,22 @@ export function getVarSize(value: any): number {
     case "object": {
       if (value instanceof HexString) {
         const size = value.byteLength;
-        return getVarSize(size) + size;
+        return getSerializedSize(size) + size;
       } else if (
         typeof value.size === "number" &&
         typeof value.serialize === "function"
       ) {
-        return getVarSize(value.size) + value.size;
+        return getSerializedSize(value.size) + value.size;
       } else if (Array.isArray(value)) {
         const array_length = value.length;
         let size = 0;
         if (array_length > 0) {
           size =
             value
-              .map((item) => getVarSize(item))
+              .map((item) => getSerializedSize(item))
               .reduce((prev, curr) => prev + curr, 0) - array_length;
         }
-        return getVarSize(array_length) + size;
+        return getSerializedSize(array_length) + size;
       }
       // do not break here so we fall through to the default
     }
--- a/packages/neon-core/src/u/misc.ts
+++ b/packages/neon-core/src/u/misc.ts
@@ -84,34 +84,36 @@ export function reverseHex(hex: string): string {
  * Calculates the byte size of any supported input following NEO's variable int format.
  */
 export function getVarSize(value: any): number {
-  if (typeof value === "object" && value instanceof HexString) {
-    const size = value.byteLength;
-    return getVarSize(size) + size;
-  } else if (typeof value === "number") {
-    if (value < 0xfd) return 1;
-    else if (value <= 0xffff) return 3;
-    else return 5;
-  } else if (Array.isArray(value)) {
-    const array_length = value.length;
-    let size = 0;
-    if (array_length > 0) {
-      const tmp = value.map((item) => {
-        // in lack of an ISerializable interface, we check it like this
-        if (
-          typeof item.size === "number" &&
-          typeof item.serialize === "function"
-        ) {
-          return item.size;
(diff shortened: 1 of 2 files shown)


snowypowers:
Possible alternatives:

1. Use `serialize(false)` which serializes the transaction excluding the witnesses. This should be sufficient for your needs and is already implemented.

1. Implement a function on Transaction that meets your needs.

Do you know where else this var size helper might be handy or would our consumers be using it?

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,39 @@ export function reverseHex(hex: string): string {
   }
   return out;
 }
+
+/**
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
+export function getVarSize(value: any): number {
```
snowypowers, line 86: Possible to type this better than just `any`?

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,39 @@ export function reverseHex(hex: string): string {
   }
   return out;
 }
+
+/**
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
+export function getVarSize(value: any): number {
+  if (typeof value === "object" && value instanceof HexString) {
```
snowypowers, line 87: Structurally, this method will read a lot better if we perform a switch case on `typeof` before narrowing it down. The object case can then be decomposed outside.

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,39 @@ export function reverseHex(hex: string): string {
...
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
+export function getVarSize(value: any): number {
+  if (typeof value === "object" && value instanceof HexString) {
+    const size = value.byteLength;
+    return getVarSize(size) + size;
+  } else if (typeof value === "number") {
+    if (value < 0xfd) return 1;
+    else if (value <= 0xffff) return 3;
+    else return 5;
```
snowypowers, line 93: The last time I implemented VarInt, there was the case of 9 bytes (see StringStream.readVarInt). Probably also requires us to throw if the number is too big?

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,39 @@ export function reverseHex(hex: string): string {
...
+  } else if (Array.isArray(value)) {
+    const array_length = value.length;
+    let size = 0;
```
snowypowers, line 96: you can just:
```
let size = value.map(item => getVarSize(item)).reduce((prev, curr) => prev + curr, 0);
```

packages/neon-core/src/u/misc.ts
ixje: I "learned" typescript 2 weeks ago, so if you have any suggestions for better typing please let me know. My IDE suggested `never`, but that came with other issues. Not sure if I should restrict it to some union of:
* hexstring
* number
* array of hex strings
* array serializable objects (specifically this one will be tricky as there is no interface afaik)

packages/neon-core/src/u/misc.ts
ixje: this part of the implementation is identical to NEO's
https://github.com/neo-project/neo/blob/ab7ec04925c1e10c9338e3deb8880e93b75ff306/src/neo/IO/Helper.cs#L99-L107

packages/neon-core/src/u/misc.ts
snowypowers: Unions would be a good start. In typescript, the types are merely a signal so a serializable interface is match as long as the underneath type has the correct method signatures.
```
interface NeonSerializable {
  serialize(): string;
}
export function getVarSize(
  value: number | string | NeonSerializable | NeonSerializable[]
): number {
  return 1;
}

getVarSize(new Witness());
getVarSize(new Transaction());
getVarSize(1);
getVarSize("hi");
```

packages/neon-core/src/u/misc.ts
ixje: 👍 
coming from python I'm not used to having a `switch` available 😅

packages/neon-core/src/u/misc.ts
ixje: good point, with a little edit it works as intended

packages/neon-core/src/u/misc.ts
snowypowers: You can read bigger VarInts than 5 bytes.
https://github.com/neo-project/neo/blob/ab7ec04925c1e10c9338e3deb8880e93b75ff306/src/neo/IO/Helper.cs#L211

ixje: what you linked is for reading byte arrays, not for reading a number

ixje:
> Possible alternatives:
> 
> 1. Use `serialize(false)` which serializes the transaction excluding the witnesses. This should be sufficient for your needs and is already implemented.
> 2. Implement a function on Transaction that meets your needs(basically copy the size() function but drop the Witness size).
> 
> Do you know where else this var size helper might be handy or would our consumers be using it?

One use-case outside of a Transaction is here: https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L412

I currently can't predict where else it will be used. It's a fair observation that this is unlikely to be used by consumers. Can we put it somewhere such that it can be used internally without exposing it to consumers?

packages/neon-core/src/u/misc.ts
ixje: shall we add `size(): number` to the `NeonSerializable` interface? 

Of course we can work around it by calling `serialize() / 2`, but I think that classes that are serializable should in general support `size()`.

ixje: Is there a smarter way of doing?
```typescript

      } else if (
        typeof value.size === "number" &&
        typeof value.serialize === "function") {
        return getVarSize(value.size) + value.size;
```
because `value instanceof NeonSerializable` doesn't work, and the remaining code complains that `NeonSerializable` doesn't have a property `serialize` 😕

snowypowers: Lets leave it there for now. I think a better way would be to formally introduce a base class for this.

Typescript transpiles to Javascript so all the interface magic goes away. Its all a facade for type checking so you cant actually attempt to apply javascript logic to types.

snowypowers:
Lets leave it in `u`. I am not opposed to exposing it since it looks like it can be useful for users who are building their own SDKs to estimate network fees.

I do suggest renaming it to `getSerializedSize` if that makes more sense. The `Var` prefix has never been useful.

ixje:
> I do suggest renaming it to `getSerializedSize` if that makes more sense. The `Var` prefix has never been useful.
I agree the NEO name is kind of poor, although I'm not entirely sure that `getSerialized` won't lead to confusion. That name would suggest it to be equal to `value.serialize().length`, but for example for the `HexString` class that is not the case
```
console.log(new Neon.u.HexString("0x1234").toArrayBuffer().length); // 2
console.log(getVarSize(new Neon.u.HexString("0x1234"))); // 3
```

snowypowers:
There isnt any other forms of serialization within the repository other than serializing for NEO. The Transaction components already use serialize and deserialize in the NEO format. I dont think it will be an issue especially if we mark everything with a interface.

ixje:
I get your point. I was trying to reason from my personal preference, where I like common names to do the same thing regardless of the project or language used.

I currently don’t have acces to my laptop, so if you want to merge now then feel free to rename the function.

snowypowers:
No rush, lets take the time to think if there's a better name. After all, naming things is one of the hardest problems in computer science :)

ixje:
> No rush, lets take the time to think if there's a better name. After all, naming things is one of the hardest problems in computer science :)

Hah! So true

I gave it some more thought and went with your suggestion anyway.

ixje:
closing in favour of #607 which is identical to this, but pulled from my personal repo such that circleci works

Issue #593, opened 2020-10-07
claim gas


lllwvlvwlll:
This feature adds helper functions to handle gas claiming

ixje:
Note: there is an undocumented RPC call `getunclaimedgas` that we can use for querying the unclaimed gas for a single address. It can be called as follows
```
{
  "jsonrpc": "2.0",
  "method": "getunclaimedgas",
  "params": ["PWYLE4NWFUqn7DRK2YhcjsRm3mp5Ts4Eod"],
  "id": 1
}
```

PR #604, opened 2020-10-08
feat(rpc): add getunclaimedgas RPC method

ixje:
fix https://github.com/CityOfZion/neon-js/issues/593

### DIFF
--- a/packages/neon-core/src/rpc/Query.ts
+++ b/packages/neon-core/src/rpc/Query.ts
@@ -133,6 +133,11 @@ export interface ValidateAddressResult {
   isvalid: boolean;
 }
 
+export interface UnclaimedGasResult {
+  unclaimed: string;
+  address: string;
+}
+
 /**
  * A Query object helps us to construct and record requests for the Neo node RPC. For each RPC endpoint, the equivalent static method is camelcased. Each Query object can only be used once.
  *
@@ -473,6 +478,15 @@ export class Query<TParams extends unknown[], TResponse> {
   ): Query<[string], ValidateAddressResult> {
     return new Query({
       method: "validateaddress",
+      params: [addr],
+    });
+  }
+
+  public static getUnclaimedGas(
+    addr: string
+  ): Query<[string], UnclaimedGasResult> {
+    return new Query({
+      method: "getunclaimedgas",
       params: [addr],
     });
   }
--- a/packages/neon-core/src/rpc/Query.ts
+++ b/packages/neon-core/src/rpc/Query.ts
@@ -133,7 +133,7 @@ export interface ValidateAddressResult {
   isvalid: boolean;
 }
 
-export interface UnclaimedGasResult {
+export interface GetUnclaimedGasResult {
   unclaimed: string;
   address: string;
 }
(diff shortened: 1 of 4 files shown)


packages/neon-core/src/rpc/Query.ts
```diff
@@ -133,6 +133,11 @@ export interface ValidateAddressResult {
   isvalid: boolean;
 }
 
+export interface UnclaimedGasResult {
```
snowypowers, line 136: Lets get the name in line with the rest of the interfaces. Top level interfaces can follow the name of the RPC call to make the link very clear.
```suggestion
export interface GetUnclaimedGasResult{

```

packages/neon-core/src/rpc/Query.ts
```diff
@@ -477,6 +482,15 @@ export class Query<TParams extends unknown[], TResponse> {
     });
   }
 
+  public static getUnclaimedGas(
```
snowypowers, line 485: A simple docstring describing the usage.

packages/neon-core/src/rpc/clients/NeoServerRpcClient.ts
```diff
@@ -258,6 +258,14 @@ export function NeoServerRpcMixin<TBase extends RpcDispatcherMixin>(
       const response = await this.execute(Query.validateAddress(addr));
       return response.isvalid;
     }
+
+    /**
+     * Get the amount of unclaimed GAS for a NEO address.
+     */
+    public async getUnclaimedGas(addr: string): Promise<number> {
+      const response = await this.execute(Query.getUnclaimedGas(addr));
+      return parseInt(response.unclaimed) / 100_000_000;
```
snowypowers, line 267: This is possible unsafe if the initial parseInt is bigger than MAX_SAFE_INTEGER. Might also want to rethink about returning a number since that too might be unsafe but should be fine in this case.

```suggestion
      return BigInteger.fromNumber(response.unclaimed).div(100_000_000).toNumber();
```

(suggestion requires adding an import I think)

snowypowers (changes requested):
Needs a unit test to ensure the data parsing works.
Needs an integration test for smoke testing.

packages/neon-core/src/rpc/clients/NeoServerRpcClient.ts
ixje: I got a "property `toNumber()` does not exist on BigInteger". So instead I changed the return value to be BigInt. Which might be more suitable anyway, given that even after dividing by the decimals factor is might (although very unlikely) still exceed MAX_SAFE_INTEGER? Let me know what you think

snowypowers: Wait I just realised we lose decimal precision in this case. We will need a better construct around this but for now Fixed8 will work. Look at Fixed8.fromRawNumber for your usage.

ixje:
> Needs a unit test to ensure the data parsing works.
> Needs an integration test for smoke testing.

Can you please point me in a direction. I'm still trying to find my way around TS and the `neon-js` library itself. Thanks

snowypowers:
neon-core/__integration__/clients/NeoServerRpcCLient.ts should be the place for integration test. We probably just wanna issue the call and ensure that we get a non-zero value (aka nothing blows up).

The unit tests is a little more tricky lets skip it for now.

packages/neon-core/src/rpc/clients/NeoServerRpcClient.ts
ixje: Done

snowypowers:
I have added a unit test with the mocks just to make sure that the parsing of the numbers work.

I have adjusted the integration test such that it will work with the integration container that we run against during normal tests. Your test will fail as the endpoint returns an error "invalid address".

snowypowers (approved):

ixje:
Thanks!

Issue #591, opened 2020-10-07
Multisig Helper Functions


lllwvlvwlll:
This feature incorporates helper functions for multisig addresses.  While not heavily used in typical applications, this functionality is critical for initial node deployment and access to tokens.

snowypowers:
Current multisig support:

- All methods within neon-core/src/wallet/multisig.ts
	- constructMultiSigVerificationScript
	- getPublicKeysFromVerificationScript
	- getSigningThresholdFromVerificationScript
	- getSignaturesFromInvocationScript
- Account.createMultiSig(signingThreshold: number, publicKeys:string[]) => Account
	Static method for creation of an Account object by assembling the verification script.
- Witness.buildMultiSig(tx:string, sigs: (string|Witness)[], acctOrVerificationScript: Account|string) => Witness
	Static method for combining the many signatures into a single Witness object.

As of now, multisig is working with a integration test in neon-core/__integration__/wallet/multisig.ts.

ixje:
I somehow think the intention here was to add [this helper function](https://github.com/neo-project/neo/blob/588cd983250f97659c6f91d7cd76ca7ff7e82990/src/neo/SmartContract/Helper.cs#L48]). Because that is required for solving a part of https://github.com/CityOfZion/neon-js/issues/581

ixje:
@kjlevitz I just realized we have to update the `function is_multisig_contract(signature_script)` signature to return the public key and signature counts if end result would be `true`. I need that information for being able to calculate the network fee ([ref](https://github.com/neo-project/neo/blob/9a160c4d249efb7364c9dbbed847fa1ded72604b/src/neo/Wallets/Wallet.cs#L417-L425))

My idea is to optionally pass an object that can be updated. So something like this
```typescript
interface SignParamsLike {
   public_key_count: number;
   signature_count: number;
}

function is_multisig_contract(signature_script: string, state?: SignParamsLike) {
    // do multisig checks
    if (state) {
        state.public_key_count = public_key_count;
        state.signature_count = signature_count;
    }
    return true;
}
```

kjlevitz:
https://github.com/kjlevitz/neon-js/blob/next/helpers/signatureVerification.ts

Current verification helper that I have been working on. If this looks good for single and multi-sig I can create a PR to next branch.

snowypowers:
[...]

PR #606, opened 2020-10-12
Helper Function for Signature checks

kjlevitz:
Created helper function to check signatures (single and multi)
Created test driver to test signature helper function

### DIFF
--- a/packages/neon-core/__tests__/rpc/Query.ts
+++ b/packages/neon-core/__tests__/rpc/Query.ts
@@ -1,5 +1,6 @@
 import Query from "../../src/rpc/Query";
-import { Transaction } from "../../src/tx";
+import { ContractParam } from "../../src/sc";
+import { Transaction, Signer } from "../../src/tx";
 
 describe("constructor", () => {
   test("RPCRequest", () => {
@@ -154,23 +155,49 @@ describe("static", () => {
     });
 
     test("multiple params", () => {
-      const params = [jest.fn(), jest.fn(), jest.fn()];
-      const result = Query.invokeFunction("hash", "method", params);
+      const inputParams = [
+        1,
+        ContractParam.integer(2),
+        { type: "Integer", value: "3" },
+      ];
+      const result = Query.invokeFunction("hash", "method", inputParams);
       expect(result.method).toEqual("invokefunction");
-      expect(result.params).toEqual(["hash", "method", params, []]);
+      expect(result.params).toEqual([
+        "hash",
+        "method",
+        [1, { type: "Integer", value: "2" }, { type: "Integer", value: "3" }],
+        [],
+      ]);
     });
 
-    test("multiple params with checkedWitnessHashes", () => {
-      const params = [jest.fn(), jest.fn(), jest.fn()];
-      const result = Query.invokeFunction("hash", "method", params, [
-        "abcdabcdabcdabcdabcd",
+    test("multiple params with signers", () => {
+      const inputParams = [
+        1,
+        ContractParam.integer(2),
+        { type: "Integer", value: "3" },
+      ];
+      const result = Query.invokeFunction("hash", "method", inputParams, [
+        new Signer({ account: "ab".repeat(20) }),
+        {
+          account: "cd".repeat(20),
+          scopes: "CalledByEntry",
+        },
       ]);
       expect(result.method).toEqual("invokefunction");
       expect(result.params).toEqual([
         "hash",
         "method",
-        params,
-        ["abcdabcdabcdabcdabcd"],
+        [1, { type: "Integer", value: "2" }, { type: "Integer", value: "3" }],
+        [
+          {
+            account: "0x" + "ab".repeat(20),
+            scopes: "FeeOnly",
+          },
+          {
+            account: "cd".repeat(20),
+            scopes: "CalledByEntry",
+          },
+        ],
       ]);
     });
   });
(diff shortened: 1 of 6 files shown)


ixje:
@snowypowers I see that the linting for this PR passes, but at the same time I see quite some inconsistency in the code style used. The following cases are all in there
```
if(something){
if (something){
if ( something){
if (something) {
```

I see `prettier` in the `eslintrc`. Shouldn't that have caught these "error"?

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
+/* 
+Helper to check if signature is a single signature
+*/
+export function isSignatureContract(signature_script){
+    const PUSHDATA1   = 12, //0x0C
```
ixje, line 5: We can use the `OpCode` constants from `import { OpCode } from "../sc";`

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
+/* 
+Helper to check if signature is a single signature
+*/
```
ixje, line 3: Try to follow the existing documentation style using 
```
/**
 * description
 */
```
```suggestion
/**
 * Check if the format of input matches that of a single signature contract
 */
```

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+/* 
+Helper to check if signatures are multi-sig
+*/
+
+interface SignParamsLike {
+    publicKeyCount: number;
+    signatureCount: number;
+ }
```
ixje, line 24: this can be removed based on https://github.com/CityOfZion/neon-js/issues/591#issuecomment-707607923

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
+/* 
+Helper to check if signature is a single signature
+*/
+export function isSignatureContract(signature_script){
```
ixje, line 4: add type for input and function return

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+/* 
+Helper to check if signatures are multi-sig
+*/
+
+interface SignParamsLike {
+    publicKeyCount: number;
+    signatureCount: number;
+ }
+
+export function isMultisigContract(signatureScript, state?: SignParamsLike){
```
ixje, line 26: add type for input and return

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+    const script = Buffer.from(signatureScript, "hex");
+    if(script.length < 43){ //First check length and return if failure to prevent indexing related errors.
+        return false;
+    }
+
+    //Hard-coded sys calls
+    const PUSHINT8 = 12, //0 //0x00
```
ixje, line 33: same as for `isSignatureContract`, re-use the existing `OpCode` constants

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+    const script = Buffer.from(signature_script, "hex");
+    if(script.length != SCRIPTLEN || script[0] != PUSHDATA1 || script[1] != 33 || script[35] != PUSHNULL || script[36] != PUSHSYSCALL || script.readUInt32LE(37) != 2014135445){
+        return false;
+    }    
+    return true;
+}
+
+/* 
+Helper to check if signatures are multi-sig
+*/
```
ixje, line 19: same comment as for the single contract version

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    if( (script.length != (i + 6)) || (script[i] != PUSHNULL) || (script[i+1] != SYSCALL) ){
+        return false;
+    }
+
+    i += 2;
+    
+    /*
+    AFEF8D13 || 2951712019
+    */
```
ixje, line 104: remove comment

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+    
+    return true;
+}
+
+/* 
+Useful notes:
+-Single Sig end instruction = 2504265080 (HEX: 95440d78)
+-Multi Sig end instruction  = 2951712019 (HEX: AFEF8D13)
+*/
```
ixje, line 121: Remove comments as this should not end up in production code

packages/neon-api/package-lock.json
```diff
@@ -0,0 +1,16 @@
+{
```
ixje, line 1: this seems like a temporary file based on your setup. I don't think this should be added

packages/neon-core/src/wallet/Account.ts
```diff
@@ -13,8 +13,9 @@ import {
   isScriptHash,
   isWIF,
 } from "./verify";
-import { InteropServiceCode } from "../sc";
+//import { InteropServiceCode } from "../sc";
```
ixje, line 16: remove comment

helpers/signatureVerification.ts
```diff
@@ -0,0 +1,121 @@
+/*
```
ixje, line 1: Please apply all comments on the functions in this file to the version in `misc.ts` and then delete this file. We don't need the functions in 2 places

ixje (changes requested):

snowypowers:
> @snowypowers I see that the linting for this PR passes, but at the same time I see quite some inconsistency in the code style used. The following cases are all in there
> 
> ```
> if(something){
> if (something){
> if ( something){
> if (something) {
> ```
> 
> I see `prettier` in the `eslintrc`. Shouldn't that have caught these "error"?

The lint rule only runs on modified files within `packages` folder. 

`helpers` folder is for runnable helpers for administrative purposes. Please move all publishable code into packages.

PR #607, opened 2020-10-13
feat(u): Add getSerializedSize

ixje:
Recreation of https://github.com/CityOfZion/neon-js/pull/603 to enable circleci

### DIFF
--- a/packages/neon-core/__tests__/u/misc.ts
+++ b/packages/neon-core/__tests__/u/misc.ts
@@ -1,6 +1,6 @@
 import * as misc from "../../src/u/misc";
 import { HexString } from "../../src/u/HexString";
-import { getVarSize } from "../../src/u/misc";
+import { getSerializedSize } from "../../src/u/misc";
 
 describe("hexXor", () => {
   test("throws if inputs of different length", () => {
@@ -58,48 +58,48 @@ class SerializableObject {
 describe("getVarSize", () => {
   test("size of HexString", () => {
     const input = HexString.fromHex("112233");
-    const result = getVarSize(input);
+    const result = getSerializedSize(input);
     expect(result).toBe(4);
   });
 
   test("size of number < 0xfd", () => {
-    const result = getVarSize(0xfc);
+    const result = getSerializedSize(0xfc);
     expect(result).toBe(1);
   });
 
   test("size of number <= 0xffff", () => {
     // lower boundary
-    let result = getVarSize(0xfd);
+    let result = getSerializedSize(0xfd);
     expect(result).toBe(3);
     // upper boundary
-    result = getVarSize(0xffff);
+    result = getSerializedSize(0xffff);
     expect(result).toBe(3);
   });
 
   test("size of number > 0xffff", () => {
-    const result = getVarSize(0xffff + 1);
+    const result = getSerializedSize(0xffff + 1);
     expect(result).toBe(5);
   });
 
   test("array of serializable objects", () => {
     const input = [new SerializableObject(), new SerializableObject()];
-    const result = getVarSize(input);
+    const result = getSerializedSize(input);
     expect(result).toBe(5);
   });
 
   test("array of hexstring objects", () => {
     const input = [HexString.fromHex("0102"), HexString.fromHex("0304")];
-    const result = getVarSize(input);
+    const result = getSerializedSize(input);
     expect(result).toBe(5);
   });
 
   test("array of unsupported objects throws", () => {
-    expect(() => getVarSize([{}, {}])).toThrow(
+    expect(() => getSerializedSize([{}, {}])).toThrow(
       "Unsupported value type: object"
     );
   });
 
   test("unsupported value throws", () => {
-    expect(() => getVarSize({})).toThrow("Unsupported value type: object");
+    expect(() => getSerializedSize({})).toThrow("Unsupported value type: object");
   });
 });
--- a/packages/neon-core/__tests__/u/misc.ts
+++ b/packages/neon-core/__tests__/u/misc.ts
@@ -88,8 +88,8 @@ describe("getSerializedSize", () => {
   });
 
   test("array of hexstring objects", () => {
-    const input = [HexString.fromHex("0102"), HexString.fromHex("0304")];
+    const input = [HexString.fromHex("0102"), HexString.fromHex("030405")];
     const result = getSerializedSize(input);
-    expect(result).toBe(5);
+    expect(result).toBe(6);
   });
 });
--- a/packages/neon-core/__tests__/u/misc.ts
+++ b/packages/neon-core/__tests__/u/misc.ts
@@ -1,6 +1,7 @@
 import * as misc from "../../src/u/misc";
 import { HexString } from "../../src/u/HexString";
 import { getSerializedSize } from "../../src/u/misc";
+import { Witness } from "../../src/tx/components";
 
 describe("hexXor", () => {
   test("throws if inputs of different length", () => {
@@ -43,18 +44,6 @@ describe("reverseHex", () => {
   });
 });
 
-class SerializableObject {
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-  // @ts-ignore
-  public get size(): number {
-    return 2;
-  }
-
-  public serialize(): string {
-    return "";
-  }
-}
-
 describe("getSerializedSize", () => {
   test("size of HexString", () => {
     const input = HexString.fromHex("112233");
@@ -82,14 +71,18 @@ describe("getSerializedSize", () => {
   });
 
   test("array of serializable objects", () => {
-    const input = [new SerializableObject(), new SerializableObject()];
-    const result = getSerializedSize(input);
-    expect(result).toBe(5);
-  });
+    const input = [
+      new Witness({
+        invocationScript: "01".repeat(256),
+        verificationScript: "03",
+      }),
+      new Witness({
+        invocationScript: "01".repeat(256),
+        verificationScript: "03",
+      }),
+    ];
 
-  test("array of hexstring objects", () => {
-    const input = [HexString.fromHex("0102"), HexString.fromHex("030405")];
     const result = getSerializedSize(input);
-    expect(result).toBe(6);
+    expect(result).toBe(523);
   });
 });
--- a/packages/neon-core/__tests__/u/misc.ts
+++ b/packages/neon-core/__tests__/u/misc.ts
@@ -55,7 +55,7 @@ class SerializableObject {
   }
 }
 
-describe("getVarSize", () => {
+describe("getSerializedSize", () => {
   test("size of HexString", () => {
     const input = HexString.fromHex("112233");
     const result = getSerializedSize(input);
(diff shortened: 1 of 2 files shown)


ixje:
No idea how to solve the lint errors. I tried
```
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
```
and it still errors. (related: https://github.com/CityOfZion/neon-js/pull/603#discussion_r502271886)

snowypowers (commented):
Comments on how to solve your lintissue.

ixje:
I've kept the `default:` case in there, because otherwise it complains about the return type. It however should be unreachable assuming the input matches `serializables` union

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,45 @@ export function reverseHex(hex: string): string {
...
+      } else if (
+        typeof value.size === "number" &&
+        typeof value.serialize === "function"
+      ) {
+        return getSerializedSize(value.size) + value.size;
+        // do not break here so we fall through to the default
+      }
+    }
+    default:
+      return 0;
```
snowypowers, line 121: Lets throw an error here in `default`. So in typescript while its typed and the transpiler will complain if we attempt it pass it something weird, during runtime, anything goes in javascript so technically anything can happen.

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,45 @@ export function reverseHex(hex: string): string {
...
+        const size = value.byteLength;
+        return getSerializedSize(size) + size;
+      } else if (Array.isArray(value)) {
+        const array_length = value.length;
+        let size = 0;
+        if (array_length > 0) {
+          size =
+            value
+              .map((item) => getSerializedSize(item))
+              .reduce((prev, curr) => prev + curr, 0) - array_length;
```
snowypowers, line 109: Why do we minus the `array_length `here?

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,45 @@ export function reverseHex(hex: string): string {
...
+      if (value < 0xfd) return 1;
+      else if (value <= 0xffff) return 3;
+      else return 5;
+    }
+    case "object": {
+      if (value instanceof HexString) {
+        const size = value.byteLength;
+        return getSerializedSize(size) + size;
+      } else if (Array.isArray(value)) {
+        const array_length = value.length;
```
snowypowers, line 103: please stick to convention and use camelCase.

packages/neon-core/__tests__/u/misc.ts
```diff
@@ -40,3 +42,54 @@ describe("reverseHex", () => {
...
+  });
+
+  test("array of serializable objects", () => {
+    const input = [new SerializableObject(), new SerializableObject()];
+    const result = getSerializedSize(input);
+    expect(result).toBe(5);
+  });
+
+  test("array of hexstring objects", () => {
+    const input = [HexString.fromHex("0102"), HexString.fromHex("0304")];
```
snowypowers, line 91: I thought this would serialize to `02020102020304`? Lets make the case for `[HexString.fromHex("0102"), HexString.fromHex("030405")]`. It does not make sense for HexString not to put down a size prefix as it is a variable sized container.

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,45 @@ export function reverseHex(hex: string): string {
...
+              .reduce((prev, curr) => prev + curr, 0) - array_length;
+        }
+        return getSerializedSize(array_length) + size;
+      } else if (
+        typeof value.size === "number" &&
+        typeof value.serialize === "function"
+      ) {
+        return getSerializedSize(value.size) + value.size;
+        // do not break here so we fall through to the default
+      }
```
snowypowers, line 118:
```suggestion
      }
        // do not break here so we fall through to the default
```
Flip the lines so the comment is clearly meant for outside the if clause.

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,45 @@ export function reverseHex(hex: string): string {
   }
   return out;
 }
+
+/**
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
```
snowypowers, line 85: lets move the docstring down right above the function.

snowypowers (changes requested):

packages/neon-core/src/u/misc.ts
ixje: ah yes that makes sense. Have to get used to having types that are not actually strict once finalized

packages/neon-core/src/u/misc.ts
ixje: because for an array we should only look at the actual length of the items in the array, whereas now using the recursion we get `length prefix` + `actual length` per item.

snowypowers: ah I believe the error might be then at calculating the size for a NeonSerializable. It should just be `return value.size`. `getSerializedSize(new SerializableObject())` should return 2?

ixje: As is, it is producing exactly what it is supposed to produce
```C#
            var witness1 = new Witness
            {
                VerificationScript = new byte[] {0x01, 0x02},
                InvocationScript = new byte[] {0x03}
            };            
            
            var witness2 = new Witness
            {
                VerificationScript = new byte[] {0x01, 0x02},
                InvocationScript = new byte[] {0x03}
            };
                

            Witness[] witnesses = { witness1, witness2};

            Console.WriteLine(witnesses.GetVarSize()); // 11
```

```javascript
const w1 = new Neon.tx.Witness({verificationScript: "0102", invocationScript: "03"});
const w2 = new Neon.tx.Witness({verificationScript: "0102", invocationScript: "03"});
const witnesses = [w1, w2];
console.log(Neon.u.getSerializedSize(witnesses)); // 11
```

snowypowers: The minus array_length works conveniently because for small objects, the prefix is only 1 byte big. It breaks down once you try to deserialize stuff larger than 0xfd.

```cs
            var largeByteArray = new byte[256];
            for (int i = 0; i < largeByteArray.Length; i++)
            {
                largeByteArray[i] = 0x01;
            }

            var witness1 = new Witness
            {
                VerificationScript = largeByteArray,
                InvocationScript = new byte[] { 0x03 }
            };

            var witness2 = new Witness
            {
                VerificationScript = largeByteArray,
                InvocationScript = new byte[] { 0x03 }
            };


            Witness[] witnesses = { witness1, witness2 };

            Console.WriteLine(witnesses.GetVarSize()); // 523
```

```js
    const input = [
      new Witness({
        invocationScript: "01".repeat(256),
        verificationScript: "03",
      }),
      new Witness({
        invocationScript: "01".repeat(256),
        verificationScript: "03",
      }),
    ];

    const result = getSerializedSize(input);
    expect(result).toBe(523); // 527
```

packages/neon-core/__tests__/u/misc.ts
ixje: The main problem is actually that we shouldn't support `HexStrings` in arrays, but we can't tell when using recursion unless we add some optional flag to the function signature to exclude `HexStrings` when called from array processing logic.

packages/neon-core/src/u/misc.ts
ixje: Good catch 👍 . 
I looked at the whole use-case again and if it is going to be used at similar places as on the Python/C# side, then we don't need to be able to parse `Serializables[]` or `NeonSerializable`, only `number | HexString | NeonSerializable[]`. I've changed the code to reflect that and updated the test cases to use your example

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,44 @@ export function reverseHex(hex: string): string {
...
+      if (value < 0xfd) return 1;
+      else if (value <= 0xffff) return 3;
+      else return 5;
+    }
+    case "object": {
+      if (value instanceof HexString) {
+        const size = value.byteLength;
+        return getSerializedSize(size) + size;
+      } else if (Array.isArray(value)) {
+        const arrayLength = value.length;
```
snowypowers, line 103: At this point, we dont have to enforce the type checks. No need for array length checks or checking if the interface conforms. The code can work fine with empty arrays and we will just throw if the caller passes something weird.
```
        const innerContentsSize = value
          .map((i) => i.size)
          .reduce((a, b) => a + b, 0);
        return getSerializedSize(value.length) + innerContentsSize;
```

packages/neon-core/src/u/misc.ts
```diff
@@ -78,3 +79,44 @@ export function reverseHex(hex: string): string {
...
+type Serializables = number | HexString | NeonSerializable[];
+/**
+ * Calculates the byte size of any supported input following NEO's variable int format.
+ */
+export function getSerializedSize(value: Serializables): number {
+  switch (typeof value) {
+    case "number": {
+      if (value < 0xfd) return 1;
+      else if (value <= 0xffff) return 3;
+      else return 5;
```
snowypowers, line 96: I do think my previous suggestion for a 9 byte var int still stands. Not clear why would you think this link is about reading byte array? The byte array method is located above and its called `ReadVarBytes`. This is called `readVarInt`.

https://github.com/neo-project/neo/blob/ab7ec04925c1e10c9338e3deb8880e93b75ff306/src/neo/IO/Helper.cs#L211

ixje: I should have originally written: "reading **from** bytearrays (==from the binary stream)".

Your `readVarInt` reads **from** a byte array (`BinaryReader` is a stream of bytes) into an `unsigned long`. It does not apply here because it always determines how many bytes to read next from the stream based on the first byte. If we applied the logic linked then
- value < 0xFD = length 1
- value == 0xFD = length 3
- value == 0xFE = length 4
- value == 0xFF = length 8

compare this to
- getSerializedSize(0xFC) = length 1 (OK)
- getSerializedSize(0xFD) = length 3 (OK)
- getSerializedSize(0xFE) = length 3 (Differs)
- getSerializedSize(0xFF) = length 3 (Differs)

We can't compare a serialized format vs a consumed format.
```C#
var num = 0xFFFF;
MemoryStream ms = new MemoryStream();
var writer = new BinaryWriter(ms);
writer.Write(num); // Note how we write a number in regular format not in varInt format
byte[] serialized = ms.ToArray();
Console.WriteLine(serialized.ToHexString()); // FFFF0000

MemoryStream ms2 = new MemoryStream(serialized);
var reader = new BinaryReader(ms2);
Console.WriteLine(Neo.IO.Helper.ReadVarInt(reader)); // exception: Unable to read beyond the end of the stream.
// because it will read the first byte (0xFF) and think it needs to read 8 more to get the integer value while there 
// are only 4 bytes in the stream in total 
```

snowypowers: An Witness array of length larger than 0xffffffff will break this code but fortunately, C# is unable to deal with it.

snowypowers (approved):

Issue #596, opened 2020-10-07
getNodes


lllwvlvwlll:
This method incorporates support for grabbing a list of public nodes for interfacing.  We can consider using the json in monitor.coz.io.

kjlevitz:
Should this just simply return an array of servers? Something along these lines:

```
  {
    protocol: 'https',
    url: 'explorer.o3node.org',
    location: 'US',
    locale: 'us',
    port: '443',
    type: 'RPC'
  },
  {
    protocol: 'https',
    url: 'rpc1.go.nspcc.ru',
    location: 'Russia',
    locale: 'ru',
    port: '10331',
    type: 'RPC'
  },
  {
    protocol: 'https',
    url: 'rpc2.go.nspcc.ru',
    location: 'Russia',
    locale: 'ru',
    port: '10331',
    type: 'RPC'
  },
  {
    protocol: 'https',
    url: 'rpc3.go.nspcc.ru',
    location: 'Canada',
    locale: 'ca',
    port: '10331',
    type: 'RPC'
  },
  {
    protocol: 'https',
    url: 'rpc4.go.nspcc.ru',
    location: 'US',
    locale: 'us',
    port: '10331',
    type: 'RPC'
  }
```

Or should it be more advanced and poll for online servers / latency?

lock9:
Hi @ixje , this is an interesting feature. We have to get info from multiple nodes to ensure we are sending our transactions to the correct node. Would it be possible to have something like getNodes, that returns a list, ordered by height & health? (not sure what I mean with health).

ixje:
What does “correct node” mean? I think we shouldn’t complicate it too much and just grab the list from dora. I think the idea behind this issue is to provide a mechanism to find a node to be used with neon-js. That’s it

PR #610, opened 2020-10-19
urls.ts now references *Server.json for return URL.

kjlevitz:
Reference #596 

urls.ts now references *Server.json for return URL.

Note: testServers.json was created with the static test URLs before they were removed.

### DIFF
--- a/helpers/mainServers.json
+++ b/helpers/mainServers.json
@@ -1,0 +1,230 @@
+  
+{
+    "name": "MainNet",
+    "pollTime": "5000",
+    "sites": [
+      {
+        "protocol": "https",
+        "url": "mainnet1.neo2.coz.io",
+        "location": "United States",
+        "port": "443",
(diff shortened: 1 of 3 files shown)


snowypowers:
Whats the purpose of putting this here in `helpers`? The current usage of `urls.ts` is for integration testing (and even that is diminished with the usage of a local docker environment). The issue you point to seems to belong in `neon-api` rather than being a mere helper.

ixje:
Kevin, a couple of remarks
* We discussed that the purpose was to scrape the data from the json files at https://github.com/CityOfZion/neo-mon because the lists there can get updated. This includes it as static data, which is not the same. 
* We also agreed to re-use the existing sorting algorithm and that the `getUrl()` could accept our dynamic list instead of being limited to the hardcoded MainNet/TestNet lists, I don't see this back in your changes.

PR #611, opened 2020-10-19
feat(u): signature verification scripts created within core

kjlevitz:
feat: 
- isSignatureContract(String) to test if provided string is a single signature.
- isMultisigContract(String) to test if provided string is a multi signature.
- isMultiSigContract has been added to \neon-js\packages\neon-core\src\u\misc.ts 

refactor!: 
- \neon-js\packages\neon-core\src\wallet\Account.ts
isMultiSig now references above isMultiSigContract function from u/mist.ts


ixje:
Why this new PR instead of updating the old one? I also don't see any of the feedback given https://github.com/CityOfZion/neon-js/pull/606 being processed

kjlevitz:
@ixje 
I apologize for removing the old branch, I was having rookie Github issues so I started everything from scratch. I will avoid this in the future. 

I made sure to address every comment that was left on the previous PR request. 

Below are my notes/replies to those comments;

https://github.com/CityOfZion/neon-js/pull/606#issuecomment-709875009
> snowypowers I see that the linting for this PR passes, but at the same time I see quite some inconsistency in the code style used. The following cases are all in there
> 
> ```
> if(something){
> if (something){
> if ( something){
> if (something) {
> ```
> 
> I see `prettier` in the `eslintrc`. Shouldn't that have caught these "error"?

This file is outside of the linting checks. This has been addressed in the lasted commit and things are now lint friendly and consistent. 

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506128536
> import { OpCode } from "../sc";

Done! Now referencing OpCode instead of local variables.


https://github.com/CityOfZion/neon-js/pull/606#discussion_r506129996
> Try to follow the existing documentation style using
> 
> ```
> /**
>  * description
>  */
> ```

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506137519
> Please apply all comments on the functions in this file to the version in `misc.ts` and then delete this file. We don't need the functions in 2 places

Both are done! Using existing comment structure. 

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506131346
> this can be removed based on [#591 (comment)](https://github.com/CityOfZion/neon-js/issues/591#issuecomment-707607923)

Done! This has been removed.

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506131925
> add type for input and function return
https://github.com/CityOfZion/neon-js/pull/606#discussion_r506132062
> add type for input and return

Done! This was required to pass lint

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506134757
https://github.com/CityOfZion/neon-js/pull/606#discussion_r506135120
https://github.com/CityOfZion/neon-js/pull/606#discussion_r506136678
> remove comment

Done!

https://github.com/CityOfZion/neon-js/pull/606#discussion_r506136200
> this seems like a temporary file based on your setup. I don't think this should be added
This will be excluded from future commits.


This should address all of the previous comments for this PR. I apologize for the untidiness of my code and will ensure it is up to standard for any future PRs.

kjlevitz:
Rebased latest next changes to this branch in an attempt to fix the unit testing errors.
Unsure exactly what is wrong, I will investigate further.

packages/neon-core/src/wallet/Account.ts
```diff
@@ -158,10 +158,7 @@ export class Account implements NeonObject<AccountJSON> {
   }
 
   public get isMultiSig(): boolean {
-    return (
-      this.contract?.script?.slice(this.contract.script.length - 8) ===
-      InteropServiceCode.NEO_CRYPTO_CHECKMULTISIGWITHECDSASECP256R1
-    );
+    return isMultiSigContract(this.contract.script);
```
ixje, line 161: the contract script is stored in base64 format which is currently not taken into account. That means `isMultiSig` wil always return false

https://github.com/CityOfZion/neon-js/blob/c6dc38a180e1651d0abe7e2afce0107a5eea023f/packages/neon-core/src/wallet/Account.ts#L30-L32

ixje (changes requested):
You seem to have changed the line endings of `misc.ts` to your local systems line ending (I'm guessing some flavour of Linux). The original was `CR LF` (Windows) and it is now `LF`. Please restore to the original line endings such that actual changes become visible, because now it looks like the whole file changed making it harder to review. If you're already using an IDE, my suggestion would be to configure it to respect the existing line endings. If you don't, now is a good time to start using one. VSCode is free and used by many. Alternatively you can try the JetBrains products (but they're paid, students get a discount though or your uni might have some deal with them). The advantage is that they'll also warn you about coding style mistakes (one of the previous comments) and often give a quick way of fixing them for the whole file.

packages/neon-core/src/wallet/Account.ts
kjlevitz: I am using VSCode oddly enough. Not sure what happened with line formatting. I can revisit that. 

As for the script, I messaged some thoughts on slack.

ixje:
superseded by #621

PR #612, opened 2020-10-25
feat(sc): A SmartContract base class

snowypowers:
A SmartContract framework for simple contract script construction. 

Before, the old framework revolves around the interface `ScriptIntent` (which is very similar to the new `ContractCall` interface). This was designed to be a simple object that SDK users can create to denote a smart contract method call. However, it was not really fully supported by the old `neon-nep5` package, which was coupled tightly with the ScriptBuilder. The `args` field was also `unknown`, leading to multiple parsing issues as there was insufficient information to verify and serialize the data.

This PR reworks `ScriptIntent` and the `neon-nep5` package for better support:

`ScriptIntent` is now superceded by `ContractCall`. With NEO3 standardizing the entry point, we can now support the (scriptHash, operation, args) inputs as a first class citizen for all contracts. Args is now typed as ContractParams which provided stable typing for us to reliably execute serialization.

The `BaseContract` class provides a base layer for constructing a simple SmartContract representation that can be used to safely construct `ContractCall` objects with their parameters parsed correctly.

The `Nep5Contract` class is an example of how the `BaseContract` class is extended and provides a clean API interface. I will further extend it to implement the NEO and GAS contracts.

Translation of a `ContractCall` into a hexstring is trivial with ScriptBuilder as all the parameters are now well defined. The `ContractCall` object can also be easily used with `invokeFunction` RPC call.

I intend to deprecate:

- ContractParam.byteArray(value, format,...args) -> ContractParam.byteArray(value). I think moving the parameter translation upwards would reduce the burden on this class. For address conversions, there is `ContractParam.hash160`. For Fixed8 conversion, we should rely on BigInteger class. I plan to remove the 2 extra arguments.

- ScriptBuilder.emitAppCall. This can be superceded by a new method to accept `ContractCall`. For now, I plan to only mark it as deprecated and visit this again after more implementation.

### DIFF
--- a/packages/neon-core/__integration__/sc/contracts/Nep5Contract.ts
+++ b/packages/neon-core/__integration__/sc/contracts/Nep5Contract.ts
@@ -1,0 +1,106 @@
+import { getIntegrationEnvUrl } from "../../../../../testHelpers";
+import { NATIVE_CONTRACTS } from "../../../src/consts";
+import { RPCClient } from "../../../src/rpc";
+import { Nep5Contract } from "../../../src/sc/contracts/Nep5Contract";
+import testWallet from "../../../__tests__/testWallet.json";
+
+let rpcClient: RPCClient;
+beforeAll(async () => {
+  const url = await getIntegrationEnvUrl();
+  rpcClient = new RPCClient(url);
+});
+
+describe("Nep5Contract", () => {
+  test("name", async () => {
+    const neoContract = new Nep5Contract(NATIVE_CONTRACTS.NEO);
+    const contractCall = neoContract.name();
+
+    const result = await rpcClient.invokeFunction(
+      contractCall.scriptHash,
+      contractCall.operation,
+      contractCall.args
+    );
+
+    expect(result.state).toBe("HALT");
+    expect(result.stack).toStrictEqual([
+      {
+        type: "ByteString",
+        value: "TkVP",
+      },
+    ]);
+  });
+
+  test("totalSupply", async () => {
+    const neoContract = new Nep5Contract(NATIVE_CONTRACTS.NEO);
+    const contractCall = neoContract.totalSupply();
+
+    const result = await rpcClient.invokeFunction(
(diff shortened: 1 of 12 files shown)


ixje:
> . For Fixed8 conversion, we should rely on BigInteger class. I plan to remove the 2 extra arguments.

Any plan to remove `Fixed8` given that it is a leftover NEO2 type?

> ScriptBuilder.emitAppCall. This can be superceded by a new method to accept ContractCall. For now, I plan to only mark it as deprecated and visit this again after more implementation.

Isn't the only change here the internal interop that is called? The argument pushing etc is still the same. If I would not take backwards compatibility in mind then all that I'd do at this point is a rename to `ScriptBuilder.emitContractCall`

snowypowers:
I will make the removal of Fixed8 the next thing on my list. 

`emitAppCall` allows for unknown parameters so its harder work with. Its a public method so I would like to remove it if possible and offer `emitContractCall` as a better solution.

ixje:
> `emitAppCall` allows for unknown parameters so its harder work with. Its a public method so I would like to remove it if possible and offer `emitContractCall` as a better solution.

Oh, so the idea is to have a strict version that sanity checks against the ABI?

snowypowers:
Yes, the BaseContract class will be in charge of doing the checks against the ABI and formatting everything correct.

ixje (approved):
The only thing that came to mind is; should we maybe do a basic validity check on the `address`(es) passed to `balanceOf` and `transfer` using `wallet.isAddress`? The rest all looks good 👍 . 

Looking forward to see what you have in mind for how to consume this contract information when building a raw transaction to be persisted.

snowypowers:
hash160 does use `isAddress` to check if it looks like an address, else it assumes its a scripthash. So in a roundabout sense, it does block invalid addresses. The error will be more of a parsing error as HexString will detect that the invalid address is not hex.

ixje:
sounds good enough

PR #614, opened 2020-10-27
feat(sc): Improve ContractParam with fromJson

snowypowers:
Rework how ContractParam construction works. Introduce fromJson and restructure the methods to have clear separation of logic. Static methods are now the preferred entry points with fromJson being the common entrypoint.

- Introduce ContractParamJson which is the direct json output coming from c#.
- Reduce constructor to type validation.
- Move parsing and transformation logic into static methods.
- Add fromJson method as a main entrypoint from external systems.
- Rework ContractParamLike to be a union of the class and json format.
- Add tests for parseEnum.

### DIFF
--- a/packages/neon-core/src/sc/ContractParam.ts
+++ b/packages/neon-core/src/sc/ContractParam.ts
@@ -15,7 +15,7 @@ export enum ContractParamType {
   PublicKey = 0x16,
   Signature = 0x17, // TODO: Implement support
 
-  Array = 0x20, // TODO: Implement support
+  Array = 0x20,
   Map = 0x22, // TODO: Implement support
 
   InteropInterface = 0x30, // TODO: Implement support
@@ -291,6 +291,7 @@ export class ContractParam implements NeonObject<ContractParamLike> {
           return ContractParam.array(...arg);
         }
         break;
+
       case ContractParamType.Boolean:
         if (
           typeof arg === "string" ||
@@ -407,6 +408,7 @@ export class ContractParam implements NeonObject<ContractParamLike> {
             );
           }
           return false;
+
         case ContractParamType.ByteArray:
         case ContractParamType.Hash160:
         case ContractParamType.Hash256:
@@ -418,6 +420,7 @@ export class ContractParam implements NeonObject<ContractParamLike> {
             return (this.value as HexString).equals(other.value);
           }
           return false;
+
         case ContractParamType.Integer:
           if (typeof other.value === "number") {
             return this.value === other.value.toString();
@@ -426,8 +429,12 @@ export class ContractParam implements NeonObject<ContractParamLike> {
             return this.value === other.value;
           }
           return false;
+
         case ContractParamType.Void:
           return true;
+
+        case ContractParamType.Boolean:
+        case ContractParamType.String:
         default:
           return this.value === other.value;
       }
--- a/packages/neon-core/src/sc/ContractParam.ts
+++ b/packages/neon-core/src/sc/ContractParam.ts
@@ -99,8 +99,9 @@ export class ContractParam implements NeonObject<ContractParamLike> {
   }
 
   /**
-   * Creates an Integer ContractParam. This is converted into an BigInteger in NeoVM. Value field will be a string.
-   * @param value - A value that can be parsed to an BigInteger. Numbers or numeric strings are accepted.
+   * Creates an Integer ContractParam. This is converted into a BigInteger in NeoVM. Value field will be a string.
+   * @param value - A value that can be parsed to a BigInteger. Numbers or numeric strings are accepted.
+
    * @example
    * ContractParam.integer(128);
    * ContractParam.integer("128");
--- a/packages/neon-core/src/sc/ContractParam.ts
+++ b/packages/neon-core/src/sc/ContractParam.ts
@@ -1,6 +1,7 @@
-import { Fixed8, reverseHex, HexString } from "../u";
+import { BigInteger, HexString } from "../u";
 import { getScriptHashFromAddress, isAddress } from "../wallet";
 import { NeonObject } from "../model";
+import { parseEnum } from "../internal";
 
 export enum ContractParamType {
   Any = 0x00,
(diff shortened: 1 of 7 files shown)


snowypowers:
@ixje this pulls out the ContractParam changes from the SmartContract PR. Please take a look.

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -64,15 +67,23 @@ export class ContractParam implements NeonObject<ContractParamLike> {
...
+    return new ContractParam({
+      type: ContractParamType.PublicKey,
+      value: value instanceof HexString ? value : HexString.fromHex(value),
+    });
   }
 
   /**
-   * Creates a Hash160 ContractParam. This is used for containing a scriptHash. Do not reverse the input if using this format.
+   * Creates a Hash160 ContractParam. This is used for containing a scriptHash. Value field will be a HexString.
+   * Do not reverse the input if using this format.
```
ixje, line 79: I find this a confusing remark, but that probably has to do with `neon-js` reversing byte arrays in many places I don't expect it to. For clarification purposes only; can you give the rules when `neon-js` does and does not expect a reversed hex string?

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -81,144 +92,235 @@ export class ContractParam implements NeonObject<ContractParamLike> {
...
+    return new ContractParam({
+      type: ContractParamType.Hash160,
+      value: HexString.fromHex(value),
+    });
   }
 
   /**
-   * Creates an Integer ContractParam. This is converted into an BigInteger in NeoVM.
+   * Creates an Integer ContractParam. This is converted into an BigInteger in NeoVM. Value field will be a string.
    * @param value - A value that can be parsed to an BigInteger. Numbers or numeric strings are accepted.
```
ixje, line 103:
```suggestion
   * Creates an Integer ContractParam. This is converted into a BigInteger in NeoVM. Value field will be a string.
   * @param value - A value that can be parsed to a BigInteger. Numbers or numeric strings are accepted.

```
just a quick grammar thing

ixje (commented):
I mostly focused on feature completion. Having said that I noticed the following;
* hash256 - which I expect to be commonly used for contracts that wish to operation on a specific block or transaction.
   * has no static creation method
   * is not supported in the normal constructor
   * is not supported in `fromJSON`
* map - I actually don't expect any contract to use this over an array type. It's going to be more expensive to push the data for no gain other than maybe readability. Regardless, if we want to be complete then
   * the same 3 as `hash256` apply here as well, additionally
   * is not supported in `toJSON` (this is known given the TODO in L: 352
* signature - never actually seen this used as a param, but for completeness
   * has no static creation method
   * is not supported in `fromJSON`
   * no explicit support in `equals`, although maybe the `default` case is sufficient?
* interop - I think we can ignore this. C# also doesn't support it in its own constructor, ToJson and FromJSON methods

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -81,144 +92,235 @@ export class ContractParam implements NeonObject<ContractParamLike> {
...
+        if (
+          typeof arg === "string" ||
+          typeof arg === "number" ||
+          typeof arg === "boolean"
+        ) {
+          return ContractParam.boolean(arg);
+        }
+        break;
+
+      case ContractParamType.ByteArray:
```
ixje, line 274: bytearray (and `signature`) are now base64 encoded
https://github.com/neo-project/neo/blob/7d3a4a2667a2c35db0b775c02082ec03e05cfa65/src/neo/SmartContract/ContractParameter.cs#L148-L151

snowypowers: The tests are all on preview3. I will upgrade the integration test container when the heads are buildable (sometimes they are not) but for now, Im sticking to preview3 compatible code which is testable.

snowypowers:
I have added Hash256 support. I decided to drop Any, Map and Signature support for the moment since they are lesser used and im not too confident how they look like.

packages/neon-core/src/sc/ContractParam.ts
snowypowers: Usually it is big endian. In this case, a Hash160's value is parsed as bigendian on C# side while a ByteArray is not. But the rule of thumb in neon-js has always been to keep moving big endian around.

ixje: Sorry it is still ambiguous for me. What do you call big endian in this context? I ask this because the NEO C# team has this crazy idea that adding a hexadecimal prefix (`0x`) suddenly changes the endianness of the data. I couldn't blame you if you follow their convention, but because it deviates from the standard I'm still confused what it means

> But the rule of thumb in neon-js has always been to keep moving big endian around.

Does this mean that you do not reverse Hash160/Hash256 types (`UInt160/UInt256` in C#) but reverse everything else?

packages/neon-core/src/sc/ContractParam.ts
ixje: Are you sure you tested this against preview3? Because this change (https://github.com/neo-project/neo/pull/1199) was introduced 12 months ago, whereas preview3 was released 3 months ago.

ixje:
> I have added Hash256 support. I decided to drop Any, Map and Signature support for the moment since they are lesser used and im not too confident how they look like.

I think that's a good call 👍

packages/neon-core/src/sc/ContractParam.ts
snowypowers: updated to use base64.

packages/neon-core/src/sc/ContractParam.ts
snowypowers: The reference case is usually the ByteArray used in C#. This is little-endian. So we extrapolate from this case to all other instances. Im mentioning the reversing because in NEO2, the convention was `ContractParam.byteArray(reverseHex(getScriptHashFromAddress(address)))`. 

In `neon-js` the `0x` prefix is usually discarded. Im aware of the convention but currently not adopting it.

What is your definition of the standard?

ixje: > What is your definition of the standard?

In general my definition is that the only time I care about endianness is when reading primitive types from a byte array where the primitive type is 2 or more bytes in size (e.g. short, int, long etc). Putting this in the NEO context and speaking from my Python project (which is a full port from C#), there are only 2 places in the whole code base where I reverse the data and they only happen for the `UInt160` and `UInt256` types.
* from string - taking a UInt160/UInt256 string representation produced by the C# code
* to string - making sure it produces the same string representation **as** the C# code (because many other tools expect this format)

Everything else internally can be said to be little endian. So my binary reader/writer classes read all data by default in little endian format. Which is why I previously mentioned that everything in `neon-js` seems to be in reversed format for me, because your representation seems to be big endian by default. :)

packages/neon-core/src/sc/ContractParam.ts
ixje: 👍

ixje (approved):

PR #616, opened 2020-11-02
feat(u): BigInteger.fromDecimal

snowypowers:
In order to transit from Fixed8, the SDK will need to be able to parse decimal numbers. The new `BigInteger` class does not support decimals correctly. This introduces methods on BigInteger to work with decimals numbers. It effectively transforms the decimal numbers into a BigInteger by shifting the decimal place.

### DIFF
--- a/packages/neon-core/__tests__/u/BigInteger.ts
+++ b/packages/neon-core/__tests__/u/BigInteger.ts
@@ -36,6 +36,19 @@ describe("twos complement", () => {
   });
 
   test.each([
+    ["0000000000000000", "0"],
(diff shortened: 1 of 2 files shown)


ixje (commented):
Not sure if it is common for `fromDecimal()` like functions, but should it accept scientific notation? e.g.  `123e-5`

Otherwise no comments

snowypowers:
I have not really seen much support for engineering/scientific notation around. Not a fan of supporting them either.

ixje:
if that is not the standard in javascript land then I'm all good

ixje (approved):

PR #627, opened 2020-11-05
feat(sc): ScriptBuilder.emitContractCall

snowypowers:
Add emitContractCall to replace emitAppCall. Also add the deprecated tag to emitAppCall.


BkunS:
Could you update [Docs](https://dojo.coz.io/neo3/neon-js/docs/api/core/sc) as well?

PR #628, opened 2020-11-09
feat(api): getTokenInfos and getTokenBalances

snowypowers:
This ports the high level facade methods from neon-nep5 into api namespace with some modifications.

- getTokenInfos returns all the relevant information that was formerly provided with totalSupply now returned as a string.
- getTokenBalances now returns an ordered array instead of a map.


ixje (commented):
`"Invoke exception: No exception returned from invoke."`
I think this is a confusing message for someone that doesn't know the internals and at first sight the first part contradicts the second. I think making it `No exception message returned.."` would clarify all of it.

Other than that looks good 👍

snowypowers:
I updated the error message and also updated getTokenInfo to getTokenInfos (usage in line with getTokenBalances).

ixje:
@snowypowers I think you forgot to push the commit that changes the error message
https://github.com/CityOfZion/neon-js/blob/e43573ccd455dc3097d1d8e3bae507c76db63942/packages/neon-api/src/api/getTokenBalances.ts#L27

Issue #624, opened 2020-11-03
Sunsetting notifications server


corollari:
I've been running the servers that serve notifications for [neo-PubSub](https://github.com/corollari/neo-PubSub) for over half a year now, but as a college student with no income I don't have the funds available to continue running them (most of the cost comes from having to maintain neo nodes for both mainnet and testnet). Due to this I'll have to sunset the service, but given that it's integrated into neon-js (the neo2 version) I think it's best to coordinate with you in order to ensure the best transition.

How do you believe it's best to proceed?

snowypowers:
We can turn it down and remove the code or potentially ask if there is any org that wishes to take over. Do you have some rough numbers?

- Amount of activity on the endpoints (eg. Amount of usage on mainnet/testnet per day or per month)
- Cost of maintaining the full stack per month.

If we can do a simple cost benefit analysis, the decision might be clearer.

corollari:
I don't run any analytics on it but I'll add some and report after some data has been collected

corollari:
Actually it turns out that the logs hold the number of subscribed clients, so I looked into it and right now that number is 8. So, overall:
- At the moment, there's 8 clients subscribed to events
- The cost of maintaining the whole thing is ~200€/mo
  - Websockets: 14$/mo. These are two heroku "hobby" dynos ($7/mo each) that distribute the events on mainnet and testnet
  - NEO nodes: ~200$/mo. These are two instances (mainnet & testnet) of the neo client running on AWS EC2 containers.

Do you have any numbers that this could be compared against?

> We can turn it down and remove the code

I see, so just remove that part of the code and release a new version on npm? Could projects that are stuck with an old version of neon-js run into problems?

snowypowers:
Paying 200 for 8 clients seems a bit too much hahaha.

Yes projects utilizing it will run into issues but the code is still there so they can host it themselves. I doubt there will be since your initial data seems to indicate little usage.

Yea we just have to remove the code. However, since the source code is available, we can leave it there and people can just host it themselves. I dont think its a big issue.

corollari:
What are your thoughts on replacing the current notification code with one compatible with neo-go's notifications? This would be a breaking change since the same information is not provided by their API. Furthermore, you mentioned earlier this year that you don't expect to release any new major versions of neon-js for neo2, so I was wondering if you believe this is a change that is worth working on and releasing or if it's better to just wait for neo3.

snowypowers:
[...]

PR #515, opened 2020-03-24
docs(api): Notification API docs

corollari:
Documentation for #514.

Pushing it as a draft PR in order to keep track of it. Feel free to close it if it's better to just open another PR when the interface for #514 is consolidated.

PR #634, opened 2020-11-21
fix(api): Remove default notification urls

corollari:
Following the discussion on #624, this PR removes the notification server urls that are used by default when the network is set to either "MainNet" or "TestNet", reverting some of the changes applied in #514 and #515.

After this PR is merged I'll make an announcement to the broader community and, after a grace period, shut down the servers.

### DIFF
--- a/docs/api/plugin/api.md
+++ b/docs/api/plugin/api.md
@@ -75,7 +75,7 @@ The available providers are `neoscan`, `neonDB` and `notifications`. The first t
 ```js
 const mainNetNeoscan = new api.neocan.instance("MainNet");
 const mainNetNeonDB = new api.neonDB.instance("MainNet");
-const mainNetNotifications = new api.notifications.instance("MainNet");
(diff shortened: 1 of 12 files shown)


corollari:
The CI failure is not related to my PR, the failing test suites are:
- `packages/neon-nep5/__integration__/main.ts`
- `packages/neon-domain/__integration__/NeoNS.ts`

snowypowers:
Can I get a url to the source code somewhere in the notifications class? That would at least point people to places where they can host it if they want to.

Dont worry about the tests I will manage it.

corollari:
I've added a reference to the source & instructions in one of the JSDoc comments in the class, is that what you were thinking or would you prefer not to expose that information to downstream consumers?

corollari:
@snowypowers thoughts?

snowypowers:
Sorry been busy. Looked good, merged.

corollari:
Thanks!

PR #635, opened 2020-11-22
fix(all): Migrate off Fixed8

snowypowers:
Migrate all API off Fixed8

With the replacement of Fixed8 with BigInteger in C#, I am also mirroring the move here. Generally, I am moving towards using integer representations instead of decimals as it is easier to work with across all contracts. It is usually trivial for the consumer to know what decimals they are working with but very difficult for the library to retain the information correctly.

There are a few important changes:

- Transaction fees are now in integer format instead of the old decimal format. Strings or BigInteger instances are now preferred.
- getUnclaimedGas also returns a integer string instead of a decimal.
- Mark fixed82num and num2fixed8 as deprecated and replace their implementations to use BigInteger.
- Update Fixed8Parser to use BigInteger under the hood. Not marked as deprecated yet.
- Mark Fixed8 as deprecated. Its likely that users will require some time to migrate off this so deprecation notice would be easier.


ixje:
🎉

PR #661, opened 2021-01-08
feat(all): Update experimental + core for preview 4

ixje:
- Update manifest to include "name"
- update ContractMethodDefinition to include "safe"
- update experimental package

Note that deploy does not work with the default RpcServer configuration. It will fail to get the system fee as the gas limit is too low. See: https://github.com/neo-project/neo-modules/issues/450


ixje:
Note that I removed `safemethods` as it moved to be a property on the methods ([ref](https://github.com/neo-project/neo/commit/f520ee59f36449088aea1ec9be5cff7aa692c484#diff-6a8fe6159fb062d52f619635b6c0e2123632e3a61acdc5f86b867da878af2e02)).  I also removed a couple of methods from the manifests as they either no longer exist, or were internal methods that were not actually exposed. For reference here's a contractstate for the NEO token obtained through RPC

```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "id": -1,
        "updatecounter": 0,
        "hash": "0x0a46e2e37c9987f570b4af253fb77e7eef0f72b6",
        "script": "DAhOZW9Ub2tlbkEa93tn",
        "manifest": {
            "name": "NeoToken",
            "groups": [],
            "supportedstandards": [
                "NEP-17"
            ],
            "abi": {
                "methods": [
                    {
                        "name": "totalSupply",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Integer",
                        "safe": true
                    },
                    {
                        "name": "setGasPerBlock",
                        "parameters": [
                            {
                                "name": "gasPerBlock",
                                "type": "Integer"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Boolean",
                        "safe": false
                    },
                    {
                        "name": "getGasPerBlock",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Integer",
                        "safe": true
                    },
                    {
                        "name": "unclaimedGas",
                        "parameters": [
                            {
                                "name": "account",
                                "type": "ByteArray"
                            },
                            {
                                "name": "end",
                                "type": "Integer"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Integer",
                        "safe": true
                    },
                    {
                        "name": "registerCandidate",
                        "parameters": [
                            {
                                "name": "pubkey",
                                "type": "ByteArray"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Boolean",
                        "safe": false
                    },
                    {
                        "name": "unregisterCandidate",
                        "parameters": [
                            {
                                "name": "pubkey",
                                "type": "ByteArray"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Boolean",
                        "safe": false
                    },
                    {
                        "name": "vote",
                        "parameters": [
                            {
                                "name": "account",
                                "type": "ByteArray"
                            },
                            {
                                "name": "voteTo",
                                "type": "ByteArray"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Boolean",
                        "safe": false
                    },
                    {
                        "name": "getCandidates",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Array",
                        "safe": true
                    },
                    {
                        "name": "getCommittee",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Array",
                        "safe": true
                    },
                    {
                        "name": "getNextBlockValidators",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Array",
                        "safe": true
                    },
                    {
                        "name": "balanceOf",
                        "parameters": [
                            {
                                "name": "account",
                                "type": "ByteArray"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Integer",
                        "safe": true
                    },
                    {
                        "name": "transfer",
                        "parameters": [
                            {
                                "name": "from",
                                "type": "ByteArray"
                            },
                            {
                                "name": "to",
                                "type": "ByteArray"
                            },
                            {
                                "name": "amount",
                                "type": "Integer"
                            },
                            {
                                "name": "data",
                                "type": "Any"
                            }
                        ],
                        "offset": 0,
                        "returntype": "Boolean",
                        "safe": false
                    },
                    {
                        "name": "symbol",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "String",
                        "safe": true
                    },
                    {
                        "name": "decimals",
                        "parameters": [],
                        "offset": 0,
                        "returntype": "Integer",
                        "safe": true
                    }
                ],
                "events": [
                    {
                        "name": "Transfer",
                        "parameters": [
                            {
                                "name": "from",
                                "type": "Hash160"
                            },
                            {
                                "name": "to",
                                "type": "Hash160"
                            },
                            {
                                "name": "amount",
                                "type": "Integer"
                            }
                        ]
                    }
                ]
            },
            "permissions": [
                {
                    "contract": "*",
                    "methods": "*"
                }
            ],
            "trusts": [],
            "extra": null
        }
    }
}
```

packages/neon-js/src/experimental/helpers.ts
```diff
@@ -1,6 +1,8 @@
 import { CONST, rpc, sc, tx, u, wallet } from "@cityofzion/neon-core";
 import { CommonConfig } from "./types";
 import { GASContract } from "./nep17";
+import { OpCode } from "@cityofzion/neon-core/lib/sc";
```
snowypowers, line 4: Please use proper imports. This is a common mistake within monorepo setups. You should be referencing sc.OpCode instead.

packages/neon-js/src/experimental/helpers.ts
```diff
@@ -208,3 +232,55 @@ export async function addFees(
...
+): Promise<string> {
+  const builder = new sc.ScriptBuilder();
+  builder.emitAppCall(CONST.NATIVE_CONTRACT_HASH.ManagementContract, "deploy", [
+    u.HexString.fromHex(u.reverseHex(NEF.toString("hex"))),
+    JSON.stringify(manifest.toJson()),
+  ]);
+
+  const transaction = new tx.Transaction();
+  transaction.script = u.HexString.fromHex(builder.build());
+  console.log(transaction.script.toString());
```
snowypowers, line 255: no console logging in general.

packages/neon-js/src/experimental/helpers.ts
```diff
@@ -208,3 +232,55 @@ export async function addFees(
...
+ */
+export async function deployContract(
+  NEF: Buffer,
+  manifest: sc.ContractManifest,
+  config: CommonConfig
+): Promise<string> {
+  const builder = new sc.ScriptBuilder();
+  builder.emitAppCall(CONST.NATIVE_CONTRACT_HASH.ManagementContract, "deploy", [
+    u.HexString.fromHex(u.reverseHex(NEF.toString("hex"))),
+    JSON.stringify(manifest.toJson()),
```
snowypowers, line 250: Is this emitting it as a Map ContractParam? Im not sure if I implemented the support correctly.

packages/neon-js/src/experimental/nep17/base.ts
```diff
@@ -1,6 +1,7 @@
 import { CONST, rpc, wallet, sc, tx, u } from "@cityofzion/neon-core";
 import { CommonConfig } from "../types";
 import { setBlockExpiry, addFees } from "../helpers";
+import { ContractParam } from "@cityofzion/neon-core/lib/sc/ContractParam";
```
snowypowers, line 4: rely on `sc.ContractParam`.

packages/neon-js/src/experimental/helpers.ts
```diff
@@ -208,3 +232,55 @@ export async function addFees(
...
 }
+
+/**
+ * Deploy a smart contract
+ * @param NEF - A smart contract in Neo executable file format. Commonly created by a NEO compiler and stored as .NEF on disk
+ * @param manifest - the manifest conresponding to the smart contract
+ * @param config -
+ */
+export async function deployContract(
+  NEF: Buffer,
```
snowypowers, line 243: Buffer isnt a common data format for neon-js. Try relaxing this constraint by allowing the web variant which is [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer). Probably can do Buffer | ArrayBuffer.

snowypowers (changes requested):
I have updated the integration docker container to have 100 Gas as invoke limit. Please try it out and add an integration test if it works.

packages/neon-js/src/experimental/helpers.ts
ixje: Yeah sorry, this is my IDE trying to be smart. I don't even notice that I couldn't call it directly, my IDE automatically adds the imports.

packages/neon-js/src/experimental/helpers.ts
ixje: good catch, left overs

packages/neon-js/src/experimental/helpers.ts
ixje: I'm guessing this emits an Array, honestly I did not look into what it creates. I just verified it is valid on the C# side.

packages/neon-js/src/experimental/helpers.ts
ixje: Might be good to look if my way of checking if the input is an ArrayBuffer type is correct.

ixje:
I'm not against trying to add an integration test, but I am against installing Docker on my machine so I need a bit of help to do this blindly as I can't query information/test locally. 
- I see there are test wallets under `/neon-core/__tests__/`, which one can I take, and specifically which account has enough balance for deploying? I need a WIF.
- I plan to create the following directory structure for the deploy test, is that ok or should I move it elsewhere? 
   - `/neon-js/__integration/experimental/`
      - helper.ts
      - contract3.nef
      - contract3.manifest.json

FYI; My idea is to adjust the below script I used to manually test it and then assert the outputs
```javascript
const Neon = require("@cityofzion/neon-js");
const fs = require("fs").promises;
const rpc_client = new Neon.rpc.RPCClient("http://127.0.0.1:10332");

const wif = "KzzBB8uAB7T5GMokeiBTEZRLJTNrokSmSEKxbpRYuYDZjAJWEuAT";
const acc = new Neon.wallet.Account(wif);
const rpc_client = new Neon.rpc.RPCClient("http://127.0.0.1:10332");

const config = {
  networkMagic: 769,
  rpcAddress: "http://127.0.0.1:10332",
  account: acc,
};

function sleep(ms) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

async function test() {
  const nef = Buffer.from(
    await fs.readFile(
      "./contract3.nef",
      null // specifying 'binary' causes extra junk bytes, because apparently it is an alias for 'latin1' *crazy*
    )
  );
  const manifest = Neon.sc.ContractManifest.fromJson(
    JSON.parse(await fs.readFile("./contract3.manifest.json"))
  );
  try {
    const contract_hash = Neon.experimental.getContractHash(
      Neon.u.HexString.fromHex(acc.scriptHash),
      nef
    );
    console.log(`Deploying contract with hash: 0x${contract_hash}`);

    console.log(await Neon.experimental.deployContract(nef, manifest, config));
    await sleep(5);
    console.log(await rpc_client.getContractState(contract_hash));
    const contract = new Neon.experimental.SmartContract(
      Neon.u.HexString.fromHex(contract_hash),
      config
    );
    console.log(await contract.testInvoke("test_func")); // expect to return Integer with value 2
  } catch (e) {
    console.log(e);
  }
}

test();
```

packages/neon-js/src/experimental/helpers.ts
snowypowers: yeah we have to test this out definitely i have not implemented support so I assume it breaks.

snowypowers:
https://github.com/snowypowers/neo-privatenet-docker/blob/v3-solonet/v3/solonet/preview4/src/setup.sh

This is the file that I used to setup the docker so you can see what addresses has what. You can just add commits and let circleci run for now. The main wallet is testWallet.json. Test code looks fine, just deploy, sleep and then test invoke it.

Alternatively push up the nef file and I can try to help but its gonna be slower.

packages/neon-js/src/experimental/helpers.ts
ixje: I'm not sure what extra you want to test regarding this emitting? I tested that the `deployContract` function works as intended. I know for sure it deserializes correctly on the C# side, because initially I
1) did not have the `u.reverseHex()` on the NEF parameter which caused the engine to fail deserializing the NEF. 
2) the manifest was missing `safe` and `name` attributes,  causing the engine to fail deserializing the manifest.
So to me that tells me that it is emitting what it needs to emit?

ixje:
The contract nef/manifest can be found in the zip attached. I'll be able to look at this again earliest tomorrow
[contract_files.zip](https://github.com/CityOfZion/neon-js/files/5801349/contract_files.zip)

ixje:
@snowypowers I need some help with the current circle-ci error related to sources files needing to be under the rootdir. This is probably some packaging stuff that I'm not familiar with. I already hardcoded the rpc client url because I was getting a similar error, but it seems to be a bigger issue related to having integration tests under the `neon-js` package? 

Previous error before hardcoding the RPC url:
```
packages/neon-js/src/__integration__/experimental/helper.ts:1:38 - error TS6059: File '/home/circleci/repo/testHelpers.ts' is not under 'rootDir' '/home/circleci/repo/packages/neon-js'. 'rootDir' is expected to contain all source files.

1 import { getIntegrationEnvUrl } from "../../../../../testHelpers";
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

packages/neon-js/src/__integration__/experimental/helper.ts:1:38 - error TS6059: File '/home/circleci/repo/testHelpers.ts' is not under 'rootDir' '/home/circleci/repo/packages/neon-js/src'. 'rootDir' is expected to contain all source files.

1 import { getIntegrationEnvUrl } from "../../../../../testHelpers";
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~

packages/neon-js/src/__integration__/experimental/helper.ts:1:38 - error TS6307: File '/home/circleci/repo/testHelpers.ts' is not listed within the file list of project '/home/circleci/repo/packages/neon-js/tsconfig.json'. Projects must list all files or use an 'include' pattern.

1 import { getIntegrationEnvUrl } from "../../../../../testHelpers";
                                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

snowypowers:
I have updated the PR with some commits. Apparently, the RPC endpoint cannot handle contract deployment invokes so I have added fees into the config as a stop gap at the moment for preview4. Else, everything works! Please take a look and then clean it up. I will chip in again after the weekend + holiday.

ixje:
this https://github.com/CityOfZion/neon-js/pull/661#pullrequestreview-565969775 made me think the RpcServer plugin gas limit was updated, but looking at this https://github.com/snowypowers/neo-privatenet-docker/blob/v3-solonet/v3/solonet/preview4/src/Plugins/RpcServer/config.json#L10 is still seems to be at 10?

If it is still `10`, then that's where the problem is with `addFees()` failing. https://github.com/neo-project/neo-modules/issues/450
I've run the above code locally against preview4 and deploy via RPC just fine. I just had to raise the gas limit as I was running out of GAS.

Other than that looks good to me, thanks! 🥇

PR #670, opened 2021-01-25
Update README.md

fabwa:


README.md
```diff
@@ -22,7 +22,7 @@ This is the JS SDK for the NEO blockchain platform. This project aims to be a li
 
 It is currently in use by [Neon](https://github.com/CityOfZion/neon-wallet/).
 
-Visit the [docs](https://cityofzion.io/neon-js) to learn how to use this library!
+Visit the [docs](https://docs.coz.io/neo3/neon-js) to learn how to use this library!
```
snowypowers, line 25: You are updating the readme for master which currently contains neo2 code. I think a better solution is to put both links here.

PR #694, opened 2021-04-02
feat(npm): Split exports into esm/cjs

snowypowers:
Reorganize build system to produce esm/cjs outputs.


ixje:
@snowypowers is this something that is still relevant now that we're in 2023 or can we close this for the time being?

PR #727, opened 2021-05-12
update magic numbers to reflect ecosystem

lllwvlvwlll:
Updates the testnet magic numbers to reflect those used in the current testnet.

### DIFF
--- a/packages/neon-core/src/consts.ts
+++ b/packages/neon-core/src/consts.ts
@@ -2,7 +2,7 @@ export const ADDR_VERSION = "35";
 
 export enum MAGIC_NUMBER {
   MainNet = 5195086,
-  TestNet = 827601742,
+  TestNet = 844378958,
   SoloNet = 1234567890,
 }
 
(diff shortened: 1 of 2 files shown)


lllwvlvwlll:
@snowypowers could you look into pushing a hotfix for RC2 that updates this?  They changed the magic numbers for RC2 (will remain the same for RC3), which is causing issues with users.  I suspect it will become a larger issues as people work on projects in the hackathon.

ixje:
Where is this `827601742` coming from? If I connect to `seed1t.neo.org:20333` I get a network id of `844378958` which matches that described in the release notes https://github.com/neo-project/neo-node/releases/tag/v3.0.0-rc2

lllwvlvwlll:
The original numbers reflected were for RC1.  I've added a commit for the RC2 numbers which should also align with the RC3 numbers.  This has been identified as a point of confusion for new users trying to interface with testnet.

snowypowers (approved):

Issue #742, opened 2021-05-28
Completing information that comes from APIs


endkeyCoder:
The properties "sysFee" and "generated" were added to "ClaimItem" class and "ClaimItemLike" Interface. The "assetHashes" property was also added to the result of the function "getBalance", for it to be possible, the property "assetHashes" was added to the "Balance" class and the property "asset_hash" to the "NeoscanBalance" interface.

motivation: Currently, the Neon Wallet app uses the Neoscan API directly to have access to the blockchain data. It was  verified that neon-js uses the same APIs to obtain the data. Knowing this, the app will no longer use the APIs (like it currently does) and instead will use neon-js functions. But some data that is returned by the API is not returned by these functions, therefore it is necessary to include that information

snowypowers:
It does seem like the implementation is converging towards the neo-cli implementation so it should be no problem supporting both new features. 

Within `neon-js`, the distinction between a token and an asset is made quite clear. I would rather keep this intact rather than go with the conflation of meaning that the official implementation went with. For this case, I prefer that the `hash` property is added to the `AssetBalance` class and then we introduce a `tokenHashes` property on the `Balance` class (which prevents modification of the old `tokens` field).

PR #745, opened 2021-06-01
feat(wallet): Add hashes to Balance/AssetBalance

endkeyCoder:
Hi @snowypowers I followed the guidelines you gave me, as soon as you can, check if you agree with the implementation, thanks for the help :)


packages/neon-core/src/wallet/Balance.ts
```diff
@@ -100,6 +101,8 @@ export class Balance {
   /** The token balances in this Balance for each token keyed by its symbol. */
   public tokens: { [sym: string]: Fixed8 };
 
+  public tokenHashes: { [sym: string]: string };
```
snowypowers, line 104: At the moment it looks like the hash is not guaranteed to exists even if the token is present within the balance. Is it possible to guarantee it?

packages/neon-api/__tests__/provider/neoCli/core.ts
```diff
@@ -136,7 +136,7 @@ describe("getClaims", () => {
   test("returns successful claims", async () => {
     const testClaim = {
       end_height: 231,
-      generated: 1464,
+      generated: 0,
```
snowypowers, line 139: Please use a non-zero unique number for `generated` and `sys_fee` since the code defaults to `0`.

packages/neon-api/__tests__/provider/neoscan/core.ts
```diff
@@ -118,6 +118,7 @@ describe("getBalance", () => {
         tokens: {
           TEST: 100,
         },
+        tokenHashes: {},
```
snowypowers, line 121: Is this possible? Will Neoscan ever return a filled token balance with no hash?

packages/neon-api/src/provider/neoscan/core.ts
```diff
@@ -120,10 +122,14 @@ export async function getBalance(
     if (b.amount > 0 && b.unspent.length > 0) {
       bal.addAsset(b.asset, {
         unspent: parseUnspent(b.unspent),
+        hash: b.asset_hash,
       } as Partial<wallet.AssetBalanceLike>);
     } else {
       bal.addToken(b.asset, b.amount);
     }
+    if (b.asset && b.asset_hash) {
```
snowypowers, line 130: This means that we are adding a token hash even if the logic is adding an asset instead. I do not think that is the correct case.

packages/neon-core/src/wallet/components/AssetBalance.ts
```diff
@@ -29,6 +31,8 @@ export class AssetBalance implements NeonObject {
     this.unconfirmed = abLike.unconfirmed
       ? abLike.unconfirmed.map((coin) => new Coin(coin))
       : [];
+
+    this.hash = abLike.hash || "";
```
snowypowers, line 35: Once again, should we be allowing this field to be unpopulated? I would say no unless we have serious issues trying to fill this field.

snowypowers (changes requested):
So far i have not seen a test case where the hash is actually populated. Please exercise the code correctly in tests.

packages/neon-core/src/wallet/Balance.ts
endkeyCoder: when querying the get_balance endpoint in both neoscan and dora, the hash is always present, the "amount" field being greater or equal than zero, and it is also present when the unspent array size is zero or greater than zero. Wouldn't that be the way to guarantee that?

packages/neon-api/__tests__/provider/neoCli/core.ts
endkeyCoder: correction applied

packages/neon-api/__tests__/provider/neoscan/core.ts
endkeyCoder: correction applied

packages/neon-api/src/provider/neoscan/core.ts
endkeyCoder: correction applied

packages/neon-core/src/wallet/components/AssetBalance.ts
endkeyCoder: correction applied

endkeyCoder:
I followed the guidelines and applied some corrections, I await further feedback, thank you @snowypowers :)

packages/neon-core/src/wallet/components/AssetBalance.ts
```diff
@@ -20,15 +24,18 @@ export class AssetBalance implements NeonObject {
   public unspent: Coin[];
   public spent: Coin[];
   public unconfirmed: Coin[];
+  public hash: string;
 
-  public constructor(abLike: Partial<AssetBalanceLike> = {}) {
+  public constructor(abLike: AssetBalanceWithHash = { hash: "nonHash" }) {
```
snowypowers, line 29: This kinds of defeats the purpose of the type system. If we are filling in a non-hash string for the user, it just means that the field itself is not guaranteed. Why not just use the old Partial<AssetBalanceLike> and throw an error if the hash field is empty.

endkeyCoder: it a good suggestion, i will do this

endkeyCoder:
I made the necessary changes, I await your review @snowypowers tks :)

packages/neon-core/__tests__/tx/calculate.ts
```diff
@@ -42,12 +48,15 @@ describe("calculateInputs", () => {
...
           unspent: [{ value: 10, txid: "ab", index: 0 }],
+          hash: NEO_ASSET_HASH,
         },
         GAS: {
           unspent: [{ value: 1.5, txid: "cd", index: 0 }],
+          hash: GAS_ASSET_HASH,
         },
         TEST: {
           unspent: [{ value: 1, txid: "ef", index: 0 }],
+          hash: NEO_ASSET_HASH,
```
snowypowers, line 59: Why does the TEST asset have the the NEO_ASSET_HASH?

packages/neon-core/__tests__/wallet/Balance.ts
```diff
@@ -23,9 +26,10 @@ describe("constructor", () => {
       assetSymbols: ["ASS"],
       tokenSymbols: ["TOK"],
       assets: {
-        ASS: {} as AssetBalanceLike,
+        ASS: { hash: NEO_ASSET_HASH } as AssetBalanceLike,
```
snowypowers, line 29: Can we not use the name NEO_ASSET_HASH if it is not? This is confusing.

packages/neon-core/__tests__/wallet/Balance.ts
```diff
@@ -66,11 +71,13 @@ describe("export", () => {
     const expected = {
       address: "address",
       net: "UnitTestNet",
+      tokenHashes: {},
```
snowypowers, line 74: Since there is the TOK token, I expect a hash for TOK here.

packages/neon-core/__tests__/wallet/components/AssetBalance.ts
```diff
@@ -1,15 +1,22 @@
-import AssetBalance from "../../../src/wallet/components/AssetBalance";
+import AssetBalance, {
+  EXCEPTION_HASH_MSG,
```
snowypowers, line 2: What is this import for?

packages/neon-core/src/wallet/verify.ts
```diff
@@ -138,3 +138,9 @@ export function isAddress(address: string): boolean {
     return false;
   }
 }
+
+export function isAssetHash(assetHash: string): boolean {
+  return (
+    (isHex(assetHash) && assetHash.length === 40) || assetHash.length === 64
```
snowypowers, line 144: I think you are mixing up assets and tokens here. Assets are UTXO styled with the 64 char hash. The 40 char hash belongs to tokens.

packages/neon-core/src/wallet/components/AssetBalance.ts
```diff
@@ -29,6 +34,11 @@ export class AssetBalance implements NeonObject {
     this.unconfirmed = abLike.unconfirmed
       ? abLike.unconfirmed.map((coin) => new Coin(coin))
       : [];
+
+    if (!abLike.hash || !isAssetHash(abLike.hash)) {
+      throw new Error(EXCEPTION_HASH_MSG);
```
snowypowers, line 39: Please inline the message. We do not need to export error strings.

packages/neon-core/__integration__/rpc/RPCClient.ts
```diff
@@ -12,7 +12,7 @@ const TESTNET_URLS = [
 ];
 
 let client: rpc.RPCClient;
-const address = "ALq7AWrhAueN6mJNqk6FHJjnsEoPRytLdW";
+const address = "AXSoNQEKjmqPBNPg5cNrHyWivfjok3Vj9D";
```
snowypowers, line 15: Is this change necessary?

endkeyCoder: yes, because the other address doesn't have a balance

packages/neon-core/src/wallet/verify.ts
endkeyCoder: ah ok, so just a "isHex" and "assetHash.lenght === 64" validation are necessary?

packages/neon-core/__tests__/tx/calculate.ts
endkeyCoder: we're going to need a valid hash to test, I thought I'd use the NEO hash as a standard hash, but it could be another one

snowypowers: This is a unit test so just generate a random hash would do.

packages/neon-core/src/wallet/verify.ts
snowypowers: yups i believe those are the only conditions.

endkeyCoder: ok but i checked the get_balance answer with other tokens like RPX or PHX and the asset_hash of those tokens is 40 characters long, wouldn't that cause a problem?

packages/neon-core/__integration__/rpc/RPCClient.ts
snowypowers: What balance are you looking at? So I mainly use that address since the private key is available in the repository for use. It would be simpler if we can continue to use the test keys instead of a custom address that the repository does not control for integration tests.

packages/neon-core/src/wallet/verify.ts
snowypowers: I think there is a misconception here. RPX and PHX are not assets. They are NEP5 tokens and do not use UTXO. Only GAS and NEO are assets so far in neo2.

packages/neon-core/__tests__/tx/calculate.ts
endkeyCoder: ok, I will do it

packages/neon-core/src/wallet/components/AssetBalance.ts
endkeyCoder: ok, I will do it

packages/neon-core/src/wallet/verify.ts
endkeyCoder: ok i really didn't notice it, I will fix the function, tks

packages/neon-core/__integration__/rpc/RPCClient.ts
endkeyCoder: I really don't need to change the address, I'll go back to the other address

endkeyCoder:
applied corrections @snowypowers thank you by your help :)

packages/neon-core/__tests__/wallet/components/AssetBalance.ts
endkeyCoder: it was a mistake, it's fixed

packages/neon-core/__tests__/wallet/Balance.ts
endkeyCoder: I agree, I changed it to DEFAULT_HASH

snowypowers (approved):

snowypowers:
please update the PR description and title to describe the change that is useful as a standalone commit.

endkeyCoder:
Below is an example of the function's return format, new fields were added to make the information more complete
```
getClaims: {
      end_height: 231,
      generated: 1337,
      n: 0,
      start_height: 48,
      sys_fee: 1337,
      txid: "dc44739e2f97743f2ed258988327560e2185ed13eec0097938eef4aea584bf04",
      unclaimed: 1464,
      value: 100000000,
    }

getBalance: {
        "net": "someUrl",
        "address":"someAddress",
        "assetSymbols": ["GAS", "NEO"],
        "assets": {
          "NEO": {
            "unspent": [{ "value": 2, "txid": "1", "index": 1 }],
            "balance": 2,
            "hash": "c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
          },
          "GAS": {
            "unspent": [{ "value": 5, "txid": "1", "index": 1 }],
            "balance": 5,
            "hash": "602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7",
          },
        },
        "tokenSymbols": ["NEO"],
        "tokens": {
          "NEO": 21,
        },
        "tokenHashes": { NEO: "c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b" },
    }
```
@snowypowers what do you think of this description?

snowypowers:
You dont have to put up an example in the description, that should be left to the tests. What I will usually put is the main changes to the API. In this PR, the core change is actually adding the tokenHashes/hash fields to the Balance/AssetBalance class. These are driven by the data available from the API.

endkeyCoder:
Added the "hash" / "tokenHashes" fields to the "Balance" / "Asset Balance" class according to the data available in the API
Fix #742

endkeyCoder:
@snowypowers that way is good?

endkeyCoder:
Hi @snowypowers , can we do merge of the code or lack something yet?

Issue #750, opened 2021-06-21
Add a “getAddressAbstracts” method for both NeoScan and Dora providers


endkeyCoder:
I need to implement a “getAddressAbstracts” for both NeoScan and Dora providers that will return address abstracts data by using the “get_address_abstracts”

PR #751, opened 2021-06-21
Add a “getAddressAbstracts” method for both NeoScan and Dora providers #750

endkeyCoder:
I need to implement a “getAddressAbstracts” for both NeoScan and Dora providers that will return address abstracts data by using the “get_address_abstracts”.

### DIFF
--- a/packages/neon-api/src/provider/apiBalancer.ts
+++ b/packages/neon-api/src/provider/apiBalancer.ts
@@ -83,15 +83,6 @@ export default class ApiBalancer implements Provider {
     return await this.loadBalance(f);
   }
 
-  public async getAddressAbstracts(
-    address: string,
-    page: number
-  ): Promise<IAddressAbstract> {
-    const f = async (p: Provider): Promise<IAddressAbstract> =>
-      await p.getAddressAbstracts(address, page);
-    return await this.loadBalance(f);
-  }
-
   /**
    * Load balances a API call according to the API switch. Selects the appropriate provider and sends the call towards it.
    * @param func - The API call to load balance function
--- a/packages/neon-api/src/provider/apiBalancer.ts
+++ b/packages/neon-api/src/provider/apiBalancer.ts
@@ -1,5 +1,5 @@
 import { logging, u, wallet } from "@cityofzion/neon-core";
-import { IAddressAbstract, PastTransaction, Provider } from "./common";
+import { AddressAbstract, PastTransaction, Provider } from "./common";
 const log = logging.default("api");
 
 export default class ApiBalancer implements Provider {
--- a/packages/neon-api/src/provider/apiBalancer.ts
+++ b/packages/neon-api/src/provider/apiBalancer.ts
@@ -1,5 +1,5 @@
 import { logging, u, wallet } from "@cityofzion/neon-core";
-import { PastTransaction, Provider } from "./common";
+import { IAddressAbstract, PastTransaction, Provider } from "./common";
 const log = logging.default("api");
 
 export default class ApiBalancer implements Provider {
@@ -80,6 +80,15 @@ export default class ApiBalancer implements Provider {
   ): Promise<PastTransaction[]> {
     const f = async (p: Provider): Promise<PastTransaction[]> =>
       await p.getTransactionHistory(address);
+    return await this.loadBalance(f);
+  }
+
+  public async getAddressAbstracts(
+    address: string,
+    page: number
+  ): Promise<IAddressAbstract> {
+    const f = async (p: Provider): Promise<IAddressAbstract> =>
+      await p.getAddressAbstracts(address, page);
     return await this.loadBalance(f);
   }
 
--- a/packages/neon-api/src/provider/common.ts
+++ b/packages/neon-api/src/provider/common.ts
@@ -13,7 +13,6 @@ export interface Provider {
   getMaxClaimAmount(address: string): Promise<u.Fixed8>;
   getHeight(): Promise<number>;
   getTransactionHistory(address: string): Promise<PastTransaction[]>;
-  getAddressAbstracts(address: string, page: number): Promise<IAddressAbstract>;
 }
 
 export interface DataProvider extends Provider {
--- a/packages/neon-api/src/provider/common.ts
+++ b/packages/neon-api/src/provider/common.ts
@@ -13,6 +13,7 @@ export interface Provider {
   getMaxClaimAmount(address: string): Promise<u.Fixed8>;
   getHeight(): Promise<number>;
   getTransactionHistory(address: string): Promise<PastTransaction[]>;
+  getAddressAbstracts(address: string, page: number): Promise<IAddressAbstract>;
 }
 
 export interface DataProvider extends Provider {
@@ -60,4 +61,22 @@ export function findGoodNodesFromHeight(
   const bestHeight = sortedNodes[0].height;
   const threshold = bestHeight - tolerance;
   return sortedNodes.filter((n) => n.height >= threshold);
+}
+
+export interface Entry {
+  txid: string;
+  time: number;
+  block_height: number;
+  asset: string;
+  amount: string;
+  address_to: string;
+  address_from: string;
+}
+
+export interface IAddressAbstract {
+  total_pages: number;
+  total_entries: number;
+  page_size: number;
+  page_number: number;
+  entries: Entry[];
 }
--- a/packages/neon-api/src/provider/common.ts
+++ b/packages/neon-api/src/provider/common.ts
@@ -72,7 +72,7 @@ export interface Entry {
   address_from: string;
 }
 
-export interface IAddressAbstract {
+export interface AddressAbstract {
   total_pages: number;
   total_entries: number;
   page_size: number;
--- a/packages/neon-api/src/provider/dora/class.ts
+++ b/packages/neon-api/src/provider/dora/class.ts
@@ -1,11 +1,12 @@
 import { logging, settings, u, wallet } from "@cityofzion/neon-core";
-import { PastTransaction, Provider } from "../common";
+import { IAddressAbstract, PastTransaction, Provider } from "../common";
 import {
   getBalance,
   getClaims,
   getHeight,
   getMaxClaimAmount,
   getRPCEndpoint,
+  getAddressAbstracts,
 } from "./core";
 
 const log = logging.default("api");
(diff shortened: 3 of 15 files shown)


snowypowers:
what does this method serve to do? At this point, it seems like you are just implementing the full API interface. Does it need to go into the common provider interface? Why not just restrict all future updates to only Dora since it is the only supported interface for all the extra calls.

endkeyCoder:
> what does this method serve to do? At this point, it seems like you are just implementing the full API interface. Does it need to go into the common provider interface? Why not just restrict all future updates to only Dora since it is the only supported interface for all the extra calls.

Hi Snowy! We need to enhance neon-js to provide this information because its currently used by Neon Wallet Mobile (we currently obtain it through our direct integration with Neoscan).

As for why we are updating the common provider interface, we understand that neoscan, as a provider for neon-js, is going to be deprecated eventually but since all methods related to it are still available to be used on neon-js, we didn't feel like it would be good to perform this enhancements just on DORA's interface. To summarize it, while neoscan is still useable through neon-js, we would be mirroring all updates we need to do on DORA's interface to Neoscan's interface as well.

If we were to bring new enhancements to neon-js in the future and notice that Neoscan interface was not accessible anymore, then we would not mirror these updates to that interface anymore.

What do you think? If you understand that we should already stop mirroring these enhancements to Neoscan interface, we can change our PRs (and future approaches) so that only Dora interface is updated.

snowypowers:
> Hi Snowy! We need to enhance neon-js to provide this information because its currently used by Neon Wallet Mobile (we currently obtain it through our direct integration with Neoscan).

Sure I can see that and it is perfectly fine.

> As for why we are updating the common provider interface, we understand that neoscan, as a provider for neon-js, is going to be deprecated eventually but since all methods related to it are still available to be used on neon-js, we didn't feel like it would be good to perform this enhancements just on DORA's interface. To summarize it, while neoscan is still useable through neon-js, we would be mirroring all updates we need to do on DORA's interface to Neoscan's interface as well.

Beyond Dora/neoscan, you are also updating the Provider interface. The `addressabstract` concept seems to be something that is specific to neoscan/dora. This does not really fit into the generic `Provider` interface as the blockchain itself does not define nor provide support for this concept. I am fine with adding the support into the specific clients but not to the generic interface. For services that depend on the address abstract, they should then just look for the specific implementation rather than try to depend on the Provider interface as that is meant for services that can use all possible combinations. I have previously accepted adjustments to the `Provider` interface for custom provider-specific methods but this is getting out of hand.

endkeyCoder:
> > Hi Snowy! We need to enhance neon-js to provide this information because its currently used by Neon Wallet Mobile (we currently obtain it through our direct integration with Neoscan).
> 
> Sure I can see that and it is perfectly fine.
> 
> > As for why we are updating the common provider interface, we understand that neoscan, as a provider for neon-js, is going to be deprecated eventually but since all methods related to it are still available to be used on neon-js, we didn't feel like it would be good to perform this enhancements just on DORA's interface. To summarize it, while neoscan is still useable through neon-js, we would be mirroring all updates we need to do on DORA's interface to Neoscan's interface as well.
> 
> Beyond Dora/neoscan, you are also updating the Provider interface. The `addressabstract` concept seems to be something that is specific to neoscan/dora. This does not really fit into the generic `Provider` interface as the blockchain itself does not define nor provide support for this concept. I am fine with adding the support into the specific clients but not to the generic interface. For services that depend on the address abstract, they should then just look for the specific implementation rather than try to depend on the Provider interface as that is meant for services that can use all possible combinations. I have previously accepted adjustments to the `Provider` interface for custom provider-specific methods but this is getting out of hand.

ok, I understood, I gonna implement the functionality of a new way, tks for your review @snowypowers

packages/neon-api/src/provider/common.ts
```diff
@@ -61,3 +61,21 @@ export function findGoodNodesFromHeight(
...
+  txid: string;
+  time: number;
+  block_height: number;
+  asset: string;
+  amount: string;
+  address_to: string;
+  address_from: string;
+}
+
+export interface IAddressAbstract {
```
snowypowers, line 75: no need to use the `I` convention found in C#. This is also the exposed interface so can we use JS conventions for all the fields (camelCase instead of snake_case).

packages/neon-api/src/provider/common.ts
```diff
@@ -61,3 +61,21 @@ export function findGoodNodesFromHeight(
   const threshold = bestHeight - tolerance;
   return sortedNodes.filter((n) => n.height >= threshold);
 }
+
+export interface Entry {
```
snowypowers, line 65: What `Entry` does this describe? Please have a name that is more specific that just entry.

packages/neon-api/src/provider/dora/class.ts
```diff
@@ -1,15 +1,24 @@
...
   getMaxClaimAmount,
   getRPCEndpoint,
+  getAddressAbstracts,
 } from "./core";
 
 const log = logging.default("api");
-export class Dora implements Provider {
+
+interface DoraProvider extends Provider {
+  getAddressAbstracts: (
```
snowypowers, line 15: I would imagine this should be singular as it returns only 1 abstract? So `getAddressAbstract`?

packages/neon-api/src/provider/dora/responses.ts
```diff
@@ -15,3 +16,10 @@ export interface DoraGetClaimableResponse {
   claimable: NeoCliClaimable[];
   unclaimed: number;
 }
+
+export type TEntry = Omit<Entry, "amount"> & { amount: number };
```
snowypowers, line 20: What does `TEntry` mean?

packages/neon-api/src/provider/dora/class.ts
```diff
@@ -46,6 +55,12 @@ export class Dora implements Provider {
   ): Promise<PastTransaction[]> {
     throw new Error("Method not implemented.");
   }
+  public getAddressAbstracts(
+    address: string,
+    page: number
```
snowypowers, line 60: Just a suggestion that you might want to have a default page number of 0 to start with?

packages/neon-api/src/provider/neoscan/class.ts
```diff
@@ -1,16 +1,24 @@
...
 
-export class Neoscan implements Provider {
+interface ProviderNeoScan extends Provider {
+  getAddressAbstracts: (
+    address: string,
+    page: number
+  ) => Promise<IAddressAbstract>;
+}
+
+export class Neoscan implements ProviderNeoScan {
```
snowypowers, line 21: Why `ProviderNeoScan` but previously `DoraProvider`? Please stick to a convention. I would prefer `NeoscanProvider`.

snowypowers (changes requested):

packages/neon-api/src/provider/neoscan/class.ts
endkeyCoder: make a sense

packages/neon-api/src/provider/dora/responses.ts
endkeyCoder: T means type and Entry means a unit of the entries field which is an array, so TENtry means the type of the entry information, basically

packages/neon-api/src/provider/common.ts
endkeyCoder: okay

packages/neon-api/src/provider/common.ts
endkeyCoder: in api response exist the entries field, which is an array, so entry represents one information these array

packages/neon-api/src/provider/dora/class.ts
endkeyCoder: I think that don't exist page zero in api response, I gonna verify but I can use number 1 like default

packages/neon-api/src/provider/dora/class.ts
endkeyCoder: the endpoint of web service is "get_address_abstracts" so I used the plural make more sense

packages/neon-api/src/provider/neoscan/class.ts
lllwvlvwlll: resolved in https://github.com/CityOfZion/neon-js/pull/751/commits/a93d3049df5306ee83254f402de0601bf2ea4107

packages/neon-api/src/provider/apiBalancer.ts
```diff
@@ -1,5 +1,5 @@
 import { logging, u, wallet } from "@cityofzion/neon-core";
-import { PastTransaction, Provider } from "./common";
```
snowypowers, line 2: I think you dont need this edit anymore

packages/neon-api/src/provider/common.ts
snowypowers: I meant something more specific. Like naming it `AddressAbstractTransactionEntry` that indicates its purpose so I do not have to try to figure out where it belongs.

packages/neon-api/src/provider/dora/class.ts
snowypowers: Ok so it looks like the `Entry` is actually the `AddressAbstract`? And your current `AddressAbstract` is just a container/header for it? If so, please resolve this by renaming `Entry` to `AddressAbstract`.

Very confusing as your raw data interface was named `DoraAddressAbstracts` but your common interface is `AddressAbstract`. Please amend that.

packages/neon-api/src/provider/dora/class.ts
snowypowers: any updates on this? This is pretty important to get right.

packages/neon-api/src/provider/dora/responses.ts
snowypowers: Please follow the convention, prefix everything with `Dora` here to indicate that its the format returned from the API endpoint.

packages/neon-api/src/provider/common.ts
```diff
@@ -61,3 +61,21 @@ export function findGoodNodesFromHeight(
...
+  time: number;
+  block_height: number;
+  asset: string;
+  amount: string;
+  address_to: string;
+  address_from: string;
+}
+
+export interface AddressAbstract {
+  total_pages: number;
```
snowypowers, line 76: All the fields in the internal/common interfaces should use camelcase for consistency. Please adhere to the convention.

snowypowers (changes requested):

packages/neon-api/src/provider/apiBalancer.ts
endkeyCoder: ok

PR #753, opened 2021-06-28
fix(u): rely on crypto-js for random bytes

### LINKED ISSUE #752: Issues with "secure-random" package using electron
```
"TypeError: r(...).randomBytes is not a function
    at http://localhost:3000/dist/bundle.js:65423:320808
    at Object.s [as default] (http://localhost:3000/dist/bundle.js:65423:321033)
    at Object.e.generateRandomArray (http://localhost:3000/dist/bundle.js:65423:419575)
    at new h (http://localhost:3000/dist/bundle.js:65423:401277)
    at e.GASContract.transfer (http://localhost:3000/dist/bundle.js:65423:9575)"
 ```
 
 The usage of the `secure-random` package is broken in the electron context within neon-js... Not exactly sure what is going on here as I am able to invoke the method it uses if I copy and paste it into my application:
 

```javascript
function nodeRandom(count, options) {
  var crypto = require('crypto')
  var buf = crypto.randomBytes(count)

  switch (options.type) {
    case 'Array':
      return [].slice.call(buf)
    case 'Buffer':
      return buf
    case 'Uint8Array':
      var arr = new Uint8Array(count)
      for (var i = 0; i < count; ++i) {
        arr[i] = buf.readUInt8(i)
      }
      return arr
    default:
      throw new Error(options.type + ' is unsupported.')
  }
}
```
It only appears to break if I use a method within neon-js that relies on it with the stack trace pointing to the `browser.js` produced by neon-js. It appears as though updates to the way that neon-js is bundled/transpiled has caused this issue

snowypowers:
Instead of `secure-random` package, rely on purely `crypto-js` to generate random bytes. Underneath the hood, both attempts to use the native crypto package in browser/node so it should work the same, the only difference being the code that checks the environment to retrieve the correct dependency. 

Possibly fix #752

### DIFF
--- a/packages/neon-core/__tests__/u/random.ts
+++ b/packages/neon-core/__tests__/u/random.ts
@@ -1,0 +1,21 @@
+import { generateRandomArray } from "../../src/u/basic/random";
+
+it("returns empty array when 0", () => {
+  const result = generateRandomArray(0);
+  expect(result.length).toBe(0);
+});
+
+it("returns correct length of bytes", () => {
+  const result = generateRandomArray(10);
+
+  expect(result.length).toBe(10);
+});
+
+it("returns numbers within 0-255", () => {
+  const randomNumbers = generateRandomArray(100);
+
+  randomNumbers.forEach((n) => {
+    expect(n).toBeLessThanOrEqual(255);
+    expect(n).toBeGreaterThanOrEqual(0);
+  });
+});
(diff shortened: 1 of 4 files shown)


comountainclimber:
@snowypowers thanks so much this seems to fully resolve the issue 🚀

PR #767, opened 2021-07-16
feat(ledger): Update for use with NEO3 ledger app

ixje:
- Tests will be updated later
- Works with https://github.com/CityOfZion/ledger-app-neo3

Sample usage
```javascript
const neonCore = require("@cityofzion/neon-core");
const ledgerPlugin = require("@cityofzion/neon-ledger").default;
const neonJs = ledgerPlugin(neonCore);

const LedgerLibrary = require("@ledgerhq/hw-transport-node-hid").default;
async function test() {
  const paths = await neonJs.ledger.getDevicePaths(LedgerLibrary);
  const ledgerInstance = await LedgerLibrary.open(paths[0]);
  console.log(await neonJs.ledger.getAppVersion(ledgerInstance));
  console.log(await neonJs.ledger.getAppName(ledgerInstance));
  console.log(await neonJs.ledger.getPublicKey(ledgerInstance, neonJs.ledger.BIP44(0)));

  const tx_string = "007b000000c8010000000000001503000000000000010000000154a64cac1b1073e662933ef3e30b007cd98d67d7000101020102";
  const network_magic = 5195086; // mainnet
  console.log(await neonJs.ledger.getSignature(ledgerInstance, tx_string, neonJs.ledger.BIP44(0), network_magic));
}

test();
```


ixje:
@lllwvlvwlll @comountainclimber

packages/neon-ledger/src/main.ts
```diff
@@ -1,48 +1,95 @@
...
+    chunk,
+    finalChunk ? 0x00 : 0x80,
     Buffer.from(msg, "hex"),
     DEFAULT_STATUSLIST
   );
 }
 
 /**
- * Appends data to the Ledger and returns the signature of the entire message that has been appended so far.
+ * Requests the public key of a requested address from the Ledger.
```
snowypowers, line 44: wrong comment?

packages/neon-ledger/src/main.ts
```diff
@@ -1,48 +1,95 @@
...
+  } catch (e) {
+    if (e.statusCode) {
+      throw evalTransportError(e as TransportStatusError);
+    }
+    throw e;
+  }
+}
+
+/**
+ * Requests the public key of a requested address from the Ledger.
```
snowypowers, line 69: update the comment.

packages/neon-ledger/src/main.ts
```diff
@@ -1,48 +1,95 @@
...
+ * @param ledger - Ledger instance
+ * @returns the application version in Major.Minor.Patch format
```
snowypowers, line 71: provide an example string.

packages/neon-ledger/src/main.ts
```diff
@@ -90,26 +137,32 @@ export async function getPublicKey(
 /**
  * Requests the device to sign a message using the NEO application.
  * @param ledger - Ledger instance
- * @param hex - message to sign as a hexstring.
+ * @param payload - message to sign as a hexstring.
  * @param bip44String - BIP44 string (40 bytes)
+ * @param networkMagic - MainNet, TestNet or custom network magic
```
snowypowers, line 142: I thought you are going to pass a string in when you mention MainNet or TestNet. I thought we moved away from the word 'magic'?

packages/neon-ledger/src/ErrorCode.ts
```diff
@@ -22,20 +37,21 @@ export interface TransportStatusError extends Error {
 export function evalTransportError(err: Error): Error {
   const transportErr = err as TransportStatusError;
   switch (transportErr.statusCode) {
-    case ErrorCode.APP_CLOSED:
```
snowypowers, line 25: Have you checked if this code still exists? I believe this code is external to the NEO app and it might still work. It is used mainly at the login screen when we detect a ledger but without the app open. Might be useful to still have this.

snowypowers (changes requested):
I dont have my ledger physically on hand at the moment but the flow looks similar. There is also a local test file named `node.test.js` which you can update and verify functionality with. `yarn test:node` should get it running.

packages/neon-ledger/src/main.ts
ixje: > I thought we moved away from the word 'magic'?

Did we? I know on the C# side they now call it `network` instead of `magic`, but on the `neon-js` a search still shows a lot of results for `magic`. Should I call it `network`?

packages/neon-ledger/src/ErrorCode.ts
ixje: Ah I didn't see that code anywhere in Ledger's new boilerplate app (what I based the new NEO app on). It seems to have changed from `0x6e00` to `0x6e01`. I moved it back in.

packages/neon-ledger/src/main.ts
snowypowers: yeah lets use `network`, gonna change it when I get some time.

packages/neon-ledger/src/main.ts
```diff
@@ -90,26 +137,32 @@ export async function getPublicKey(
...
-    const response = await finalizeDataForSignature(
+    const response = await sendDataToSign(
       ledger,
-      chunks[chunks.length - 1]
+      chunks[chunks.length - 1],
+      2 + chunks.length,
+      true
     );
-    if (response.readUIntBE(0, 2) === ErrorCode.VALID_STATUS) {
+    if (response.readUIntBE(0, 2) === StatusWord.OK) {
```
snowypowers, line 165: On a side note, what is actually the correct status that should be returned here? My understanding was that the ledger will keep returning OK as it accepts new data for signature and if we still see OK at this point, it means something weird happened. If we know the correct return value, we can check for that instead.

snowypowers (approved):

packages/neon-ledger/src/main.ts
ixje: Every `apdu` send to the device will get at least 2 bytes of response in the form of a status word. `0x9000` just means the apdu was processed as expected, otherwise we get some other status word like `0x6e01` for `APP_CLOSED`. It can also reply with "data + status word". It is up to us to decide how the protocol works. What you describe could very well have been the behaviour of the old app. 

For the new app there is no such thing as "OK" means we can receive more data. The new "protocol" for signing an app is
-> send bip44 path
<- `0x9000`
-> send network magic
<- `0x9000`
-> send unsigned data part of the tx
<- receive signature + `0x9000`

Having said that I realize that piece of code is no longer applicable. I removed it

snowypowers: upon failure, I suppose the ledger bridge should throw an error which would invoke the try-catch. Can you add a comment explaining the latest edit? Probably something along the lines of `Expected signature + 2 status bytes to be returned here upon completion`

snowypowers:
Do you want to try and update the dependencies too? The latest version of `@ledgerhq/hw-transport` is pretty far head of the current one. Else we can run with this for now.

snowypowers:
I am going to merge this first. You can tinker with the package versions later in a separate PR.

PR #794, opened 2021-10-14
fix for getversion change in Neo 3.0.3

hal0x2328:
Neo 3.0.3 introduced a breaking change to the `getversion` RPC method. The `network` value was moved one level down under `protocol`, which broke the `Facade` api in neon-js.


snowypowers:
can you open this from a fork instead? The tests only run when its a PR from a fork.

PR #853, opened 2022-06-29
fix(sc): Remove GetRandom from fixed interop price list

ixje:
following https://github.com/neo-project/neo/pull/2749


snowypowers:
are these PRs targeting v3.1 or v3.3+? If its 3.3+, can we point them to dev?

ixje:
They're for parity with v3.3.0+

PR #857, opened 2022-07-15
fix(rpc): fix wrong type for ApplicationLogJson

ixje:
Reported by Discord user @vzin777

the `state` field of notifications are StackItems not Contract parameters. 
image

image


snowypowers (commented):
Remove the yarn.lock changes and we are good.

PR #865, opened 2022-08-12
fix(experimental): Apply prioritisation fee

### LINKED ISSUE #864: CommonConfig's prioritisationFee is not used and might lead to confusion
https://github.com/CityOfZion/neon-js/blob/804b20a34872be3d6ae70bc036bdf367fc50b280/packages/neon-js/src/experimental/types.ts#L5

ixje:
Fix #864

### DIFF
--- a/packages/neon-js/src/experimental/helpers.ts
+++ b/packages/neon-js/src/experimental/helpers.ts
@@ -199,6 +199,12 @@ export async function addFees(
   transaction: tx.Transaction,
   config: CommonConfig
 ): Promise<void> {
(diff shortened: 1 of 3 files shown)


packages/neon-js/src/experimental/helpers.ts
```diff
@@ -224,6 +224,18 @@ export async function addFees(
     );
   }
 
+  if (config.networkFeeOverride && config.prioritisationFee) {
```
snowypowers, line 227: can we do the precondition checks at the start of the method so we can do an early exit.

snowypowers (approved):

Issue #856, opened 2022-07-11
deprecate calculateNetworkFee in experimental


ixje:
following hardfork stuff in neo-cli. should point to using `invokeScript` instead or add a replacement helper

ixje:
completed in v5.2.0 release

PR #866, opened 2022-08-12
feat(api): smartCalculateNetworkFee

### LINKED ISSUE #849: Smart RPC calculateNetworkFee or restructure and remove deprecation note
https://github.com/CityOfZion/neon-js/blob/bb298756d4bcb6edeb3c545816bc9c92bceae3c3/packages/neon-api/src/api/calculateNetworkFee.ts#L11-L13

The `calculateNetworkFee` function is marked deprecated and suggests using the "recently" introduced [calculateNetworkFee](https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/calculatenetworkfee.html) RPC method. I think this is a good suggestion because recently there are some hard fork things happening in the code that affect interop prices. However, using the RPC method isn't straight forward at all. Let me elaborate

In order for the `calculateNetworkFee` RPC method to work correctly one needs a transaction with sufficient witnesses. The problem is that witnesses aren't added until we sign, and we can't sign until we have the network fee because that is part of the signed data. So we we have a chicken-egg problem.

The solution is to add temporary witnesses (with at least a valid `verificationScript`) prior to calling the RPC method, then remove them after obtaining the fee and prior to signing. This is internal knowledge that we can't expect most users to have. This trick is actually what the existing code is doing but locally. 

Now comes the question how to make this accessible to the end-users, I see 2 options
1. remove this from the existing function https://github.com/CityOfZion/neon-js/blob/bb298756d4bcb6edeb3c545816bc9c92bceae3c3/packages/neon-api/src/api/calculateNetworkFee.ts#L44-L64, replace it with a RPC call to `calculateNetWorkFee` passing `txClone` as input and remove the deprecated note. This means the function needs to become `async` and it leaves you with an RPC call that is arguably unusable without the internal knowledge described above and pushed that responsibility to the user.
2. Make `calculateNetworkFee` intelligent by adding the temporary witnesses there.

-edit- my vote would go for option 2, what's your take @snowypowers ?

ixje:
fix #849 & #856

### DIFF
--- a/packages/neon-js/src/experimental/helpers.ts
+++ b/packages/neon-js/src/experimental/helpers.ts
@@ -194,9 +194,14 @@ export async function setBlockExpiry(
 /**
  * Add system and network fees to a transaction.
  * Validates that the source Account has sufficient balance
+ *
+ * Note: Witnesses must be present on the transaction. If no witnesses are
+ * present a temporary single signature account witness will be used for
+ * fee calculation. For fee calculation using a multi signature account you
+ * must add the witness yourself. See TransactionBuilder.addEmptyWitnesses()
+ * for reference how this could be done.
  * @param transaction - the transaction to add network and system fees to
  * @param config -
- * @param token_decimals -
  */
 export async function addFees(
   transaction: tx.Transaction,
--- a/packages/neon-js/src/experimental/helpers.ts
+++ b/packages/neon-js/src/experimental/helpers.ts
@@ -1,12 +1,15 @@
 import { CONST, rpc, sc, tx, u, wallet } from "@cityofzion/neon-core";
 import { CommonConfig } from "./types";
 import { GASContract } from "./nep17";
+import { smartCalculateNetworkFee } from "@cityofzion/neon-api";
 
 /**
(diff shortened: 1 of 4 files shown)


packages/neon-js/src/experimental/helpers.ts
```diff
@@ -217,11 +220,20 @@ export async function addFees(
...
-    transaction.networkFee = await calculateNetworkFee(
-      transaction,
-      config.account,
-      config
-    );
+    const rpcClient = new rpc.RPCClient(config.rpcAddress);
+    const txClone = new tx.Transaction(transaction);
+
+    if (txClone.witnesses.length < 1) {
+      txClone.addWitness(
```
snowypowers, line 227: this assumes a simple signer (no multisig) which is fine. Is this constraint mentioned anywhere?

snowypowers (approved):

packages/neon-js/src/experimental/helpers.ts
ixje: Good point, it is not. How does the updated doc string work for you?

snowypowers: LGTM.

PR #868, opened 2022-08-15
fix(rpc): ContractParam identification

### LINKED ISSUE #850: Strange Behavior when using `sc.ContractParam`
Hello, I notice a problem with some methods of `sc.ContractParam`, they can't be directly used with `invokeFunction.

## How to Reproduce
Making a simple request like the following:
```ts
const params = [sc.ContractParam.hash160(address)]; // <------- using the method directly

const res = await new rpc.RPCClient(rpcAddress).invokeFunction(
      '0xb2c806468effacd2951b913e724e4c3f8506a357',
      'balanceOf',
      params
    )
```

### Result
Results in a request body like the following:
(This result was retrieved by monitoring the Chrome Dev Tools, Network tab)
```json
{
   "params":[
      "0xb2c806468effacd2951b913e724e4c3f8506a357",
      "balanceOf",
      [
         {
            "type":20,
            "value":{
                // <------- empty object
            }
         }
      ],
      [
         
      ]
   ],
   "jsonrpc":"2.0",
   "id":1234,
   "method":"invokefunction"
}
```
The value is an empty object.

## Workaround
I can workaround this problem by using `toJson` method:
```ts
const params = [sc.ContractParam.hash160(address).toJson()];
```

### Result
Resulting in a working request body:
```json
{
   "params":[
      "0xb2c806468effacd2951b913e724e4c3f8506a357",
      "balanceOf",
      [
         {
            "type":"Hash160",
            "value":"cc776527da4a34b80f411b0ccb9dffddb38523ae" // <-------
         }
      ],
      [
         
      ]
   ],
   "jsonrpc":"2.0",
   "id":1234,
   "method":"invokefunction"
}
```

I had to use `toJson` with the following methods:
- hash160
- byteArray
- string (not always)
- integer
- publicKey

Is this expected? Is it happening to you too?

ixje:
close #850 

This is a possible solution, perhaps there is a better one.

### DIFF
--- a/packages/neon-core/src/sc/ContractParam.ts
+++ b/packages/neon-core/src/sc/ContractParam.ts
@@ -513,6 +513,7 @@ export function isContractParam(value: any): value is ContractParam {
     value.type !== undefined &&
     value.type in ContractParamType &&
     value.value !== null &&
-    value.value !== undefined
+    value.value !== undefined &&
+    typeof value.toJson === "function"
   );
 }
--- a/packages/neon-core/src/sc/ContractParam.ts
+++ b/packages/neon-core/src/sc/ContractParam.ts
@@ -502,3 +502,17 @@ export function likeContractParam(
     cp.value !== undefined
   );
 }
+
+// eslint-disable-next-line
+export function isContractParam(value: any): value is ContractParam {
+  if (value instanceof ContractParam) {
+    return true;
+  }
+
+  return (
+    value.type !== undefined &&
+    value.type in ContractParamType &&
+    value.value !== null &&
+    value.value !== undefined
+  );
+}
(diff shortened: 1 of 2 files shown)


packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -502,3 +502,17 @@ export function likeContractParam(
     cp.value !== undefined
   );
 }
+
+// eslint-disable-next-line
```
ixje, line 506: otherwise I get a 
```
Unexpected any. Specify a different type
```
if I use `unknown`  then it errors on the `value` at lines 513-516 with the message
```
Object is of type 'unknown'
```

ixje:
superseded by #869

PR #869, opened 2022-08-23
fix(docs): add import note

### LINKED ISSUE #850: Strange Behavior when using `sc.ContractParam`

ixje:
closes #850 as per https://github.com/CityOfZion/neon-js/issues/850#issuecomment-1221930903
supersedes #868

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,11 @@ console.log(Neon);
 var acct = Neon.create.account("NKuyBkoGdZZSLyPbJEetheRhMjeznFZszf");
 ```
 
+> **Note**
+> mixing import `neon-core` and `neon-js` (the latter which includes `neon-core`) in the same project can lead to undesired
+> behaviour ([example](https://github.com/CityOfZion/neon-js/issues/850)). Unless all you need is contained in `neon-core`, use
+> `neon-js`.
+
 # Contributing
 
 Please refer to [`CONTRIBUTING`](./CONTRIBUTING.md) for development practices.
--- a/README.md
+++ b/README.md
@@ -58,10 +58,11 @@ console.log(Neon);
 var acct = Neon.create.account("NKuyBkoGdZZSLyPbJEetheRhMjeznFZszf");
 ```
 
+
 > **Note**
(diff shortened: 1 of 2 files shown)


snowypowers (commented):
I think we can be more direct with the recommendation. Let's lead with the TLDR and an active voice.

I suggest:

```
For most use-cases, we recommend `neon-js`.
Do not use `neon-js` and `neon-core`  in the same project.  The classes are not cross-package compatible. See https://github.com/CityOfZion/neon-js/issues/850.
```

snowypowers (approved):

PR #870, opened 2022-08-23
fix(docs): update core docs

ixje:
Yesterday somebody pointed out that the core docs where not completely up to date either. I updated all of them except for the `tx.md` one (more on this later). The goal was to make all samples and text valid again. I updated wallet addresses and transaction id, hashes and such to point valid Neo3 values

As for `tx.md` I'm not sure what would be valuable content. The majority of it now describes the different transactions we had in Neo2. Stuff like this are no longer a thing
```
Neon.create.claimTx(...args);
```

 The only functions we could describe are `addWitness`, `addSigner` and maybe `(de)serialize`. `addAttribute` exists but is obsolete NEO2 stuff.

### DIFF
--- a/docs/api/core/tx.md
+++ b/docs/api/core/tx.md
@@ -22,14 +22,6 @@ processed into a block by the consensus nodes.
 // We can use this serializedTx string and send it through sendrawtransaction RPC call.
 const serializedTx = transaction.serialize();
 ```
-
-For deserialization, we have the `Transaction` static class:
-
-```js
-const transaction = tx.Transaction.deserialize(hexstring);
-console.log(transaction instanceof tx.Transaction); // true
-```
-
 ---
 
 ## Components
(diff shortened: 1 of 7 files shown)


ixje:
Took a stab at updating `tx.md`. Tx building now happens through the `TransactionBuilder`, which is actually part of `neon-api`  not `neon-core` so kind of falls outside of this documentation

docs/api/core/tx.md
```diff
@@ -6,154 +6,54 @@ title: Transactions
...
-
-const hash = tx.hash; // Store the hash so we can use it to query a block explorer.
-
-// Now we can use this serializedTx string and send it through sendrawtransaction RPC call.
-const serializedTx = tx.serialize();
+// We can use this serializedTx string and send it through sendrawtransaction RPC call.
+const serializedTx = transaction.serialize();
 ```
 
 For deserialization, we have the `Transaction` static class:
```
snowypowers, line 26: this is also outdated and can be removed.

Issue #876, opened 2022-11-08
BooleanWitnessCondition has no serialize method override


xie-yue:
in [BooleanWitnessCondition](https://github.com/CityOfZion/neon-js/blob/dev/packages/neon-core/src/tx/components/WitnessCondition.ts#L132)

there is no serialize method override, so `expression` didn't work.
for deserialize read byte for `expression`, next step `deserializeAttributes` throws en error `11 not found in TxAttrUsage!`

related
https://github.com/CityOfZion/neon-js/issues/872
@ixje

ixje:
Sorry to see that. Hopefully the above PR fixes all remaining issues with WitnessConditions. 

As for `deserializeAttributes`. That's a separate issue. Transaction attributes are something only committee members can set and as a result hardly ever seen. The current implementation is actually still based on NEO2

ixje:
@xie-yue fyi; https://www.npmjs.com/package/@cityofzion/neon-js/v/5.2.2

xie-yue:
works great! thx! @ixje

PR #878, opened 2022-11-08
fix(tx): support NEO3 transaction attributes

ixje:
The current code is still legacy Neo2. The transaction attributes have changed (and reduced) significantly. There are only 2 now
1. HighPriority
2. OracleResponse

* `High priority` can only be set by consensus committee members ([ref1](https://github.com/neo-project/neo/blob/77ee2cc5b6ea371efdf3be506b173c6304b0fc01/src/Neo/Network/P2P/Payloads/HighPriorityAttribute.cs#L38))
*  `OracleReponse` only by an entity voted into the network as OracleService by the consensus committee ([ref2](https://github.com/neo-project/neo/blob/77ee2cc5b6ea371efdf3be506b173c6304b0fc01/src/Neo/Network/P2P/Payloads/OracleResponse.cs#L103)). 
 
For these reasons I removed the ability to add attributes to a TX. Instead I only added support for (de)serialization because it seems liked the user [here](https://github.com/CityOfZion/neon-js/issues/876) was processing a TX with such attributes (likely an OracleResponse). Because I work so irregularly with JS/TS I took the style from `WitnessCondition`.

### DIFF
--- a/packages/neon-core/src/tx/components/TransactionAttribute.ts
+++ b/packages/neon-core/src/tx/components/TransactionAttribute.ts
@@ -1,139 +1,216 @@
-import { num2hexstring, num2VarInt, StringStream, HexString } from "../../u";
-import { TxAttrUsage } from "./txAttrUsage";
-import { NeonObject } from "../../model";
+import {
+  num2VarInt,
+  StringStream,
+  HexString,
+  NeonSerializable,
+  reverseHex,
+  base642hex,
+} from "../../u";
+import { parseEnum } from "../../internal";
 
-const maxTransactionAttributeSize = 252;
+export enum TransactionAttributeType {
+  HighPriority = 0x1,
+  OracleResponse = 0x11,
+}
+
+export enum OracleResponseCode {
+  // Indicates that the request has been successfully completed.
+  Success = 0x00,
+  // Indicates that the protocol of the request is not supported.
+  ProtocolNotSupported = 0x10,
+  // Indicates that the oracle nodes cannot reach a consensus on the result of the request.
+  ConsensusUnreachable = 0x12,
+  // Indicates that the requested Uri does not exist.
+  NotFound = 0x14,
+  // Indicates that the request was not completed within the specified time.
+  Timeout = 0x16,
+  // Indicates that there is no permission to request the resource.
+  Forbidden = 0x18,
+  // Indicates that the data for the response is too large.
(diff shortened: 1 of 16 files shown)


melanke:
Task linked: [CU-2ve3xje update transaction attributes](https://app.clickup.com/t/2ve3xje)

packages/neon-core/src/tx/components/TransactionAttribute.ts
```diff
@@ -1,139 +1,216 @@
...
-  // Base64-encoded data
-  data: string;
+export interface OracleResponseTransactionAttributeJson {
+  type: string;
+  // request id
+  id: number;
+  // response code
+  code: string;
+  // base64 encoded result for the request
+  result: string | HexString;
```
snowypowers, line 50: if it is base64 encoded, it cannot be a HexString.

packages/neon-core/src/tx/components/TransactionAttribute.ts
```diff
@@ -1,139 +1,216 @@
...
+  type: string;
 }
 
-export interface TransactionAttributeJson {
-  // Name of TransactionAttributeUsage in English
-  usage: string;
-  // Base64-encoded data
-  data: string;
+export interface OracleResponseTransactionAttributeJson {
+  type: string;
```
snowypowers, line 44: in typescript, you can enforce this to be a static string "OracleResponse"  to be stricter. (Similarly for HighPriority)

ixje: I think the change is what you asked for

packages/neon-core/src/tx/components/TransactionAttribute.ts
```diff
@@ -1,139 +1,215 @@
...
-    if (TxAttrUsage.Url !== type) {
-      throw new Error(`${type} not found in TxAttrUsage!`);
-    }
-  }
+export type TransactionAttributeJson =
+  | HighPriorityTransactionAttributeJson
+  | OracleResponseTransactionAttributeJson;
 
-  return type as TxAttrUsage;
+export interface TransactionAttributeLike {
```
snowypowers, line 56: is the `Like` interfaces being used? I cant seem to find them anywhere?

snowypowers (approved):

ixje:
They’re used with the export() function. Which Transaction.export() calls on the attributes.

PR #892, opened 2023-04-14
fix(wallet): resolve slow scrypt encryption/decryption with react

### LINKED ISSUE #891: 5.3.0 is much slower on key encryption/decryption than 4.8.0
Simple code like
```
			const acc = new wallet.Account();
			const wall = Neon.create.wallet({ name: 'some' });
			wall.addAccount(acc);
			wall.encrypt(0, password)
```
or
```
			const account = new wallet.Account(parsedWallet.accounts[0].key);
			account.decrypt(password)
```
is 100% portable between 4.8.0 (Legacy) and 5.3.0 (N3). And it effectively does the same NEP-2 crypto behind the scenes. But we've noticed that using 4.8.0 it takes ~2-3 seconds per operation, while using 5.3.0 it takes ~15s. And we're not sure why it happens this way. Scrypt parameters (that can influence) are the same in both versions.

CC @mike-petrov.

ixje:
fix #891 

I can't explain the difference between packages, I just know there is a speed difference despite both being pure-js. 
I've considered `hash-wasm` as alternative to `ethereum-cryptography`, but it was actually slower and required more changes.


snowypowers:
Is there really a speed difference? Just looking at the test runtimes (this [PR](https://app.circleci.com/pipelines/github/CityOfZion/neon-js/1087/workflows/e92499fc-fe50-41af-8275-cb1bb3406e39) and an older [run](https://app.circleci.com/pipelines/github/CityOfZion/neon-js/1084/workflows/df3d7f72-f3c4-422c-8fd6-41f369422045), the runtimes are very similar.

I am not opposed to using ethereum's package since its more modern & supported but calling it a fix without any stats is not ideal. Do you have any runtime numbers? How about pitching this as a "modernization" instead of a true speed fix?

ixje:
The test cases indeed do not show much if any difference. The difference is there when used in combination with React. If you check the issue you’ll see a discussion including some code playgrounds where we get very high delays. With this package, in the specific problematic scenario, the speed difference goes from ~13 seconds to sub 1 second.On 15 Apr 2023, at 06:00, Yak Jun Xiang ***@***.***> wrote:﻿
Is there really a speed difference? Just looking at the test runtimes (this PR and an older run, the runtimes are very similar.
I am not opposed to using ethereum's package since its more modern & supported but calling it a fix without any stats is not ideal. Do you have any runtime numbers? How about pitching this as a "modernization" instead of a true speed fix?

snowypowers (approved):

PR #895, opened 2023-05-23
CU-864ekh1aw - new Account of NeonJS needs to accept '0x' and remove it if necessary

raulduartep:
This PR is important to avoid problems when passing `0x` on the scripthash to create a new Account.
It seems unecessary at first but this constructor is being used by other libraries that tries to convert different kinds of input to an Address or ScriptHash.

### DIFF
--- a/packages/neon-core/src/u/basic/hex.ts
+++ b/packages/neon-core/src/u/basic/hex.ts
@@ -1,6 +1,17 @@
 const hexRegex = /^([0-9A-Fa-f]{2})*$/;
 
 /**
+ * Format a hexstring to a supported format.
+ */
+export function remove0xPrefix(str: string): string {
+  if (str.startsWith("0x")) {
+    str = str.substring(2);
+  }
+
+  return str;
+}
+
+/**
  * Checks if input is a hexstring. Empty string is considered a hexstring.
  */
 export function isHex(str: string): boolean {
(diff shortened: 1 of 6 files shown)


lllwvlvwlll:
Task linked: [CU-864ekh1aw new Account of NeonJS needs to accept "0x" and remove it if necessary](https://app.clickup.com/t/864ekh1aw)

packages/neon-core/src/u/basic/hex.ts
```diff
@@ -1,5 +1,16 @@
 const hexRegex = /^([0-9A-Fa-f]{2})*$/;
 
+/**
+ * Format a hexstring to a supported format.
+ */
+export function formatHex(str: string): string {
```
melanke, line 6: I think this name is too generic and doesn't actually describes what the method is doing. I think it could be named `remove0xPrefix`, `strip0xPrefix`, `removeLeading0x`, something like this.

melanke, line 6: I think this name is too generic and doesn't actually describes what the method is doing. I think it could be named `remove0xPrefix`, `strip0xPrefix`, `removeLeading0x`, something like this.

melanke (changes requested):

melanke:
@raulduartep actually I think this PR is not following neon-js directives:
1) The commit name doesn't follow the standard
2) You need to send a PR from a fork, like simplitech/neon-js
3) You are sending to `master` but it should be `dev`

PR #896, opened 2023-05-23
feat: accept scriptHash with 0x prefix in Account constructor

raulduartep:
This PR is important to avoid problems when passing 0x on the scripthash to create a new Account.
It seems unecessary at first but this constructor is being used by other libraries that tries to convert different kinds of input to an Address or ScriptHash.

### DIFF
--- a/docs/changelog/v5.md
+++ b/docs/changelog/v5.md
@@ -5,6 +5,13 @@ sidebar_label: Latest (v5)
 sidebar_position: 1
 ---
 
+# 5.4.1
+
+- wallet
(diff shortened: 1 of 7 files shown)


lllwvlvwlll:
Task linked: [CU-864ekh1aw new Account of NeonJS needs to accept "0x" and remove it if necessary](https://app.clickup.com/t/864ekh1aw)

melanke (approved):

snowypowers (approved):
you might wanna add to the changelog too to track the changes.

PR #898, opened 2023-07-06
feat(sc): add support to map parameter

raulduartep:


lllwvlvwlll:
Task linked: [CU-864ej5xh6 On ContractParam of neon-js, implement Map type](https://app.clickup.com/t/864ej5xh6)

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -435,18 +501,31 @@ export class ContractParam implements NeonObject<ContractParamLike> {
...
           }
           return false;
 
+        case ContractParamType.Map:
+          if (
+            Array.isArray(this.value) &&
+            Array.isArray(other.value) &&
+            this.value.length === other.value.length
+          ) {
+            return (this.value as ContractParamMap).every(
```
snowypowers, line 521: this assumes that the map is ordered. Is this the case? (u have no tests for this at the moment)

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -16,16 +16,31 @@ export enum ContractParamType {
...
 
 export interface ContractParamJson {
   type: string;
-  value?: string | boolean | number | ContractParamJson[] | null;
+  value?:
+    | string
+    | boolean
+    | number
+    | ContractParamJson[]
+    | ContractParamMapLike
```
snowypowers, line 42: The json representation should only include json stuff. So you will need a `ContractParamMapJson` type here. (The usage of the *Json types is at the serialization boundary so we should be assured that the input is always plain json)

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -16,16 +16,31 @@ export enum ContractParamType {
...
 
   Void = 0xff,
 }
+export type ContractParamMapLike = {
+  key: ContractParamLike;
+  value: ContractParamLike;
+}[];
+
+export type ContractParamMap = {
+  key: ContractParam;
```
snowypowers, line 31: on a side note, does neo actually allow anything to be the key? Can I put a map as the key?

ixje: No you can't. Only `string`, `bool` and `int` can be used as key type (what they call `PrimitiveType` in the VM).

packages/neon-core/src/sc/ContractParam.ts
ixje: a `map` on the VM side is ordered [ref](https://github.com/neo-project/neo-vm/blob/5b0a39811b34abacab1273f3ee5a9a9f7e52ac7b/src/Neo.VM/Types/Map.cs#L21)

packages/neon-core/src/sc/ContractParam.ts
raulduartep: @snowypowers Should I create a validation for it ?

snowypowers: If we can constraint the type, then yes please. Else, at least some comments would help out.

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -16,22 +16,44 @@ export enum ContractParamType {
...
   Array = 0x20,
-  Map = 0x22, // TODO: Implement support
+  Map = 0x22,
 
   InteropInterface = 0x30, // TODO: Implement support
 
   Void = 0xff,
 }
 
+export type ContractParamMap = {
```
snowypowers, line 26: Propose to rename this to `ContractParamMapEntry`.  `ContractParamMap` is too close to the type `ContractParam.Map` and causes some confusion.

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -16,22 +16,44 @@ export enum ContractParamType {
...
+
 export interface ContractParamJson {
   type: string;
-  value?: string | boolean | number | ContractParamJson[] | null;
+  value?:
+    | string
+    | boolean
+    | number
+    | ContractParamJson[]
+    | ContractParamMapJson
```
snowypowers, line 48: I believe this should be `ContractParamMapJson[]` (or `ContractParamMapEntryJson[]` if renamed)

packages/neon-core/src/sc/ContractParam.ts
```diff
@@ -16,22 +16,44 @@ export enum ContractParamType {
...
-  Map = 0x22, // TODO: Implement support
+  Map = 0x22,
 
   InteropInterface = 0x30, // TODO: Implement support
 
   Void = 0xff,
 }
 
+export type ContractParamMap = {
+  key: ContractParam;
```
snowypowers, line 27: Do add a comment here to constraint the type.

snowypowers (approved):

PR #922, opened 2023-11-22
fix(deps): update crypto-js to 4.2.0

luc10921:
There is already a [chore](https://github.com/CityOfZion/neon-js/pull/919) that is kinda doing this but since the docs and `types` on `package.json` should also be updated, I was instructed to create another PR with these other changes.

As to why this PR is changing only `crypto-js`, it's because it's the only dependency that is raising a severe vulnerability.
image


melanke (approved):
I didn't understand the lint problem, but the PR looks ok to me.

snowypowers (approved):

PR #924, opened 2024-01-22
feat(rpc):  add findStorage method

luc10921:
I was not able to run the integration tests locally. So I'm trying to use GitHub actions to run it. However, if it doesn't use the TestNet, then the test will fail (it's trying to find the AccountPrefix on NEO, and expecting it to have at least 51 accounts).


lock9 (approved):

packages/neon-core/__tests__/rpc/Query.ts
snowypowers, line 1: please include a test that verifies the optional start param (when we pass 2 params, the default for start is 0)

luc10921: Oh, I should've asked this before: the `findstorage` RPC call does not accept 2 parameters, it needs the third one. But is it ok to have this default behavior of adding the third one as zero on neon-js?

snowypowers: yeah i think that is fine.

packages/neon-core/src/rpc/Query.ts
```diff
@@ -524,6 +532,23 @@ export class Query<TParams extends JsonRpcParams, TResponse> {
     });
   }
 
+  /**
+   * This Query returns the keys and values stored with a specific key prefix under a specific contract.
+   * @param scriptHash - contract script hash
+   * @param searchPrefix - prefix to search for
+   * @param start - start location.
```
snowypowers, line 539: I think `index` is a better word than location.

packages/neon-core/src/rpc/clients/NeoServerRpcClient.ts
```diff
@@ -182,6 +183,26 @@ export function NeoServerRpcMixin<TBase extends RpcDispatcherMixin>(
       return response;
     }
 
+    /**
+     * Finds all values with a given prefix key in the storage of a contract address.
+     *
+     * @param scriptHash - contract script hash
+     * @param searchPrefix - prefix to search for, can be empty string to return all keys in storage
+     * @param start - start location, the page size is limited to 50 items, if more than 50 values exist on the
```
snowypowers, line 191: similar to above comment about start index.

snowypowers (approved):

snowypowers:
i updated dev with a newer integration test container which should contain the newer neo version needed to get findStorage working. Please rebase and see if it works.

packages/neon-core/__integration__/rpc/clients/NeoServerRpcClient.ts
```diff
@@ -226,6 +226,52 @@ describe("NeoServerRpcClient", () => {
...
+
+      const findTotalSupplyPrefixWithDefaultArgResult =
+        await client.findStorage(contractHash, "0b");
+
+      // This find will only get the totalSupply of NEO. Should be safe and static for usage.
+      expect(findTotalSupplyPrefixResult.truncated).toBe(false);
+      expect(findTotalSupplyPrefixResult.next).toBe(1);
+      expect(findTotalSupplyPrefixResult.results.length).toBe(1);
+      expect(findTotalSupplyPrefixResult.results[0].key).toBe("Cw==");
+      expect(findTotalSupplyPrefixResult.results[1].key).toBe("AOH1BQ==");
```
snowypowers, line 245: `TypeError: Cannot read properties of undefined (reading 'key')`

Now that I look closer, results.length is 1 so you cannot retrieve the [1] of it.

packages/neon-core/__integration__/rpc/clients/NeoServerRpcClient.ts
```diff
@@ -226,6 +226,52 @@ describe("NeoServerRpcClient", () => {
...
+    });
+
+    test("find NEOs balance storage", async () => {
+      // This find will only get a list of accounts on NEOs storage. Will only be used to check the truncated and next values.
+      const findAccountPrefix0Result = await client.findStorage(
+        contractHash,
+        "14",
+        0
+      );
+      expect(findAccountPrefix0Result.truncated).toBe(true);
```
snowypowers, line 259: unfortunately, within the integration test container, the number of accounts with NEO is <10 so this will naturally fail. I dont think there is another way to do this unless i craft a custom script to spread the neo around 51 addresses.

luc10921: I see, I thought it was synced to the mainnet or testnet. I don't think I'll be able to check a storage where I get more than 50 results. I'll rewrite the test to expect less than 50

snowypowers (approved):

Issue #928, opened 2024-04-08
Invalid checksum error when deploy contract(version3.6.2)


Tobeyw:
When we deployed the 3.6.2 version contract with`neonjs`, an` invalid checksum` error occurred. 
I found that the latest version of `n3 compiler 3.6.2` deleted the `MaxScriptLength` field of the `nefFile file`, which changed the calculation method of `chenkSum`. I guess it may be related to this.
https://github.com/neo-project/neo/commit/efa86592d080ebfde685bc87a26177c5bed5431f#diff-e5a9d7d849aceff46a01663d9e6246d09c4c82e49af60f21dc879d05c40481ee

![image](https://github.com/CityOfZion/neon-js/assets/38420178/08d211af-b6d9-4457-9788-b78138011813)

ixje:
The format of the NEF hasn't changed with that commit. It seems to only change the allowed size of the `script` field from `512*1024` (as defined by `MaxScriptLength`) to `ushort.max * 2` (as defined by `ExecutionEngineLimits.Default.MaxItemSize`).

I'll try out the new 3.6.2. compiler and see if I can reproduce

luc10921:
I tried running the `fromBuffer` method using my NEF file generated with the 3.6.2 compiler and it worked.
```ts
    const myNef = sc.NEF.fromBuffer(
      readFileSync(
        "path\\to\\nef\\file"
      )
    )
```

Do you know which neon-js version you are using?

Tobeyw:
`@cityofzion/neon-js": "^5.5.1",`

ixje:
Can you please share the code snippet used that caused the error and also what you now use that works? The checksum in the file seems correct, so this is a strange case and we need a bit more information to investigate.

Tobeyw:
This is the code I deploy the contract with neon-js. [neonjs](https://github.com/Tobeyw/TestN3-neon/blob/1d13f9fe02fc2a58770d78ff69a6947aec318a2f/deployContract.js#L21)

This is the code I deploy the contract with [C#](https://github.com/Tobeyw/TestN3/blob/b0ce2e9d23194e953143f128349fa48321e0bc99/main.cs#L114) and it worked.

ixje:
[...]

PR #929, opened 2024-04-10
fix(sc): fix NEF's compiler field deserialization when occupying 64 bytes

luc10921:
Fixes https://github.com/CityOfZion/neon-js/issues/928


ixje (approved):
lgtm!

PR #931, opened 2024-11-11
Added Static Method to Account Class for Key Validation

digimbyte:
Introduced a `validateKey` static method in the `Account` class that validates a string input and returns its type if valid, or `void`/`undefined` if invalid. This method streamlines validation by reducing the need for try/catch blocks and complex conditional checks, providing a clear and direct approach for assessing the health status of a string input.

### DIFF
--- a/packages/neon-core/src/wallet/Account.ts
+++ b/packages/neon-core/src/wallet/Account.ts
@@ -29,7 +29,7 @@ export enum KeyType {
   Address = "Address",
   WIF = "WIF",
   NEP2 = "NEP2",
-  unknown = "",
+  Unknown = "",
 }
 
 export interface AccountJSON {
@@ -112,7 +112,7 @@ export class Account implements NeonObject<AccountJSON> {
       case isNEP2(str): return KeyType.NEP2;
       default:
         // returning empty string enables falsly checks
-        return KeyType.unknown;
+        return KeyType.Unknown;
     }
   }
 
--- a/packages/neon-core/src/wallet/Account.ts
+++ b/packages/neon-core/src/wallet/Account.ts
@@ -21,6 +21,16 @@ import { NeonObject } from "../model";
 
 const log = logger("wallet");
 
+export enum KeyType {
+  PrivateKey = "PrivateKey",
+  PublicKeyUnencoded = "PublicKeyUnencoded",
+  PublicKeyEncoded = "PublicKeyEncoded",
+  ScriptHash = "ScriptHash",
+  Address = "Address",
+  WIF = "WIF",
+  NEP2 = "NEP2",
+}
+
 export interface AccountJSON {
   /** Base58 encoded string */
   address: string;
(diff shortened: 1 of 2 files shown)


packages/neon-core/src/wallet/Account.ts
```diff
@@ -88,6 +88,20 @@ export class Account implements NeonObject<AccountJSON> {
     });
   }
 
+  public static validateKey(str: string): string | void {
```
snowypowers, line 91: Correct me if I am wrong but conventionally, validate returns a bool. If the objective is to also provide some information about the string, may I suggest the name `identifyString `instead.

snowypowers, line 91: Instead of string | void, consider either:

- narrowing the return type to the specific strings (eg. 'PrivateKey' | 'EncodedPublicKey' | ...)
- Adding an enum to consts and referencing that

packages/neon-core/src/wallet/Account.ts
```diff
@@ -88,6 +88,20 @@ export class Account implements NeonObject<AccountJSON> {
     });
   }
 
+  public static validateKey(str: string): string | void {
+    switch (true) {
+      case isPrivateKey(str): return "PrivateKey";
+      case isPublicKey(str, false): return "PublicKey_Decoded";
```
snowypowers, line 94: Prefer to follow the current name [PublicKeyUnencoded ](https://github.com/CityOfZion/neon-js/blob/ffbf31e45c9b69ffc02e9cb7b0aad746b67400d0/packages/neon-core/src/wallet/core.ts#L49C20-L49C37)

Similarly for `PublicKeyEncoded`.

snowypowers (commented):
unit tests?

digimbyte:
@snowypowers thanks for looking at it, I wrote it mid project while juggling other code
upgraded to enums per requested and fixed a few missing elements.
should resolve with falsely values

packages/neon-core/src/wallet/Account.ts
digimbyte: can do both with a falsely check with an empty string

packages/neon-core/src/wallet/Account.ts
```diff
@@ -21,6 +21,17 @@ import { NeonObject } from "../model";
...
 
+export enum KeyType {
+  PrivateKey = "PrivateKey",
+  PublicKeyUnencoded = "PublicKeyUnencoded",
+  PublicKeyEncoded = "PublicKeyEncoded",
+  ScriptHash = "ScriptHash",
+  Address = "Address",
+  WIF = "WIF",
+  NEP2 = "NEP2",
+  unknown = "",
```
snowypowers, line 32: small nit: enum capitalisation. Should be Unknown

snowypowers:
i have no other issues except the lack of unit tests though the code is indeed pretty straightforward.

digimbyte:
I'm not too familiar with the current unit testing implemented. but I fixed the case for the Unknown.

Issue #954, opened 2026-04-23
Replace or mitigate elliptic in neon-core 5.x due GHSA-848j-6mx2-7j84


aboimpinto:
HushForge is using `@cityofzion/neon-js@5.8.1` for Neo N3 script/app integration, and
Dependabot/npm audit is flagging the current `elliptic` dependency chain:

```text
@cityofzion/neon-js@5.8.1
  -> @cityofzion/neon-core@5.8.1
  -> elliptic@6.6.1
```

Advisory:

```text
GHSA-848j-6mx2-7j84
Elliptic Uses a Cryptographic Primitive with a Risky Implementation
Affected range: elliptic <= 6.6.1
```

`npm audit fix` suggests downgrading `@cityofzion/neon-js` to `3.11.9`, but that is a major
downgrade from 5.x and is not a safe migration path for us because we rely on the current v5 API
surface.

## What I checked

I cloned the current `CityOfZion/neon-js` source and checked the direct `elliptic` usage.

The dependency is declared here:

```text
packages/neon-core/package.json
```

The direct crypto wrapper is concentrated here:

```text
packages/neon-core/src/u/basic/curve.ts
```

That wrapper is used by wallet key/signing flows, including:

```text
packages/neon-core/src/wallet/signing.ts
packages/neon-core/src/wallet/core.ts
```

I also ran the focused baseline tests before changing anything:

```powershell
npx jest packages/neon-core/__tests__/u/basic/curve.ts packages/neon-core/__tests__/wallet/signing.ts --runInBand
```

Baseline result:

```text
Test Suites: 2 passed, 2 total
Tests:       30 passed, 30 total
```

## Compatibility concerns

The direct source area is small, but the behavior is security-sensitive:

- `curve.ts` supports both `SECP256R1` / `p256` and `SECP256K1`.
- Neo N3 wallet signing uses `SECP256R1`.
- Existing tests cover compressed/uncompressed public keys, decode, sign, and verify.
- Existing tests also cover the optional explicit nonce `k` behavior.
- `wallet/signing.ts` includes exact expected signatures using `k = "0400"`.

That means this probably should not be treated as a simple dependency bump.

## Shim experiment

Based on community feedback, I also tested aliasing `elliptic` to the compatibility shim:

```json
{
  "dependencies": {
    "elliptic": "npm:@soatok/elliptic-to-noble@9999.0.0"
  }
}
```

This did remove the production `elliptic` audit chain in the local experiment, but the focused
[...]

PR #955, opened 2026-04-23
fix(neon-core): replace elliptic with noble curves

aboimpinto:
## Summary

Draft PR for #954.

This replaces Neon-Core's direct `elliptic` usage with `@noble/curves` while preserving the existing Neon-JS 5.x curve/signing API surface.

Changes:

- remove `elliptic` and `@types/elliptic` from `@cityofzion/neon-core`;
- add direct `@noble/curves@1.9.7` dependency;
- replace `packages/neon-core/src/u/basic/curve.ts` implementation with Noble-backed P-256 and secp256k1 support;
- preserve compressed/uncompressed public key behavior;
- preserve verification with `lowS: false` to match existing behavior;
- preserve the existing explicit `k` parameter through a small manual ECDSA nonce path;
- add `docs/elliptic-replacement.md` with investigation notes, assumptions, and baseline evidence.

## Assumptions to review

The main assumption is around the existing public `k` parameter on `EllipticCurve.sign(message, privateKey, k)`.

Noble's public signing API does not expose the same arbitrary caller-provided nonce `k` behavior, while Neon-JS has tests that rely on deterministic signatures with `k`, including exact wallet signing vectors using `k = "0400"`.

To avoid changing public behavior silently, this PR preserves the `k` behavior with a small manual ECDSA path. Please review this part carefully and advise if you prefer removing/deprecating/changing that API instead.

## Validation

```text
npx jest packages/neon-core/__tests__/u/basic/curve.ts packages/neon-core/__tests__/wallet/signing.ts --runInBand
=> 2 suites passed, 30 tests passed

npm run build --workspace packages/neon-core
=> passed

npm run lint --workspace packages/neon-core
=> passed

npx jest /packages/neon-core/__tests__/.* --runInBand
=> 43 suites passed, 760 tests passed, 10 skipped

npm audit --omit=dev
=> no elliptic finding remains; one unrelated lodash advisory remains

npm run dist:prod --workspace packages/neon-core
=> webpack compiled successfully with existing-style bundle-size warnings

npm run dist:prod --workspace packages/neon-js
=> webpack compiled successfully with existing-style bundle-size warnings
```

## Known local command caveat

```text
npm test --workspace packages/neon-core -- --runInBand
```

This command fails before running tests because `packages/neon-core/jest.config.js` attempts to mutate an undefined transform entry. Running the same Neon-Core tests from root Jest works and passed as shown above.


snowypowers:
I see the overheads of trying to maintain `k` for a deterministic signature is pretty ridiculous. Lets forgo it (its only for testing too, in real life we will never expect k to be set).

aboimpinto:
Agreed. I removed the manual custom-k ECDSA implementation and updated the tests away from exact legacy nonce vectors. The optional k parameter remains accepted/deprecated for source compatibility but is ignored by the Noble-backed signing path.

packages/neon-core/src/u/basic/curve.ts
```diff
@@ -17,51 +42,32 @@ export interface EcdsaSignature {
...
-        r: signature.r.toString("hex", 32),
-        s: signature.s.toString("hex", 32),
-      };
-    }
     const signature = this.curve.sign(
-      Buffer.from(message, "hex"),
-      Buffer.from(privateKey, "hex"),
+      hexToBytes(message),
+      hexToBytes(privateKey),
+      { lowS: false },
```
snowypowers: is there a strong reason to allow generating malleable signatures? I understand verify can be more lax but I dont see a reason for producing a weaker signature than needed.

packages/neon-core/src/u/basic/curve.ts
```diff
@@ -1,12 +1,37 @@
...
 import { Buffer } from "buffer";
 
 export interface EcdsaSignature {
   r: string;
   s: string;
 }
 
+type NobleCurve = typeof p256;
+
+function bytesToHex(bytes: Uint8Array): string {
```
snowypowers: https://github.com/paulmillr/noble-curves/tree/v1#utils-useful-utilities there are utils that we can take advantage of.

packages/neon-core/src/u/basic/curve.ts
```diff
@@ -93,23 +96,27 @@ export class EllipticCurve {
...
    * @param publicKey - 33 byte hexstring starting with 02 or 03.
    */
   public decodePublicKey(publicKey: string): string {
-    const publicKeyBuffer = Buffer.from(publicKey, "hex");
-    return this.curve
-      .keyFromPublic(publicKeyBuffer, "hex")
-      .getPublic()
-      .encode("hex", false);
+    return bytesToHex(
+      this.curve.ProjectivePoint.fromHex(hexToBytes(publicKey)).toRawBytes(
```
snowypowers: Not sure why we have to `fromHex(hexToBytes(publicKey))` when the public key should be a hexstring? And if there is a fromHex, is there a `toHex`? Would be weird not to provide that functionality.

packages/neon-core/src/u/basic/curve.ts
```diff
@@ -17,51 +42,32 @@ export interface EcdsaSignature {
...
-      };
-    }
     const signature = this.curve.sign(
-      Buffer.from(message, "hex"),
-      Buffer.from(privateKey, "hex"),
+      hexToBytes(message),
+      hexToBytes(privateKey),
+      { lowS: false },
     );
+    const bytesLength = this.getScalarBytesLength();
```
snowypowers: do we have an reason for this? I believe 32 just works unless we are trying to service a multitude of curves?

packages/neon-core/src/u/basic/curve.ts
aboimpinto: Agreed. I changed signing to use Noble’s normal low-S behavior, so we no longer produce malleable signatures.

Verification stays permissive with `lowS: false` so existing/legacy high-S signatures can still be accepted

packages/neon-core/src/u/basic/curve.ts
aboimpinto: Good point. I replaced the local Buffer-based hex helpers with Noble’s `bytesToHex` / `hexToBytes` utilities.

packages/neon-core/src/u/basic/curve.ts
aboimpinto: Agreed, that conversion was unnecessary. `fromHex` can take the hex string directly, and Noble points expose `toHex`.

Updated to:

  ```ts
  return this.curve.ProjectivePoint.fromHex(publicKey).toHex(false);

packages/neon-core/src/u/basic/curve.ts
aboimpinto: You’re right. Both supported curves here are 32-byte curves, so the dynamic scalar-byte helper was unnecessary.

I removed it and now split Noble’s compact signature hex into 32-byte `r` and `s` components.

aboimpinto:
**Optional k Parameter**
I removed the custom nonce implementation path.
The optional `k` argument is still accepted at the TypeScript boundary for source compatibility, but it is ignored and signing uses Noble’s normal path

snowypowers:
I think we are good once we remove the doc? I dont think it belongs. With this I think we can bump a minor version for release. I glanced through the rest quickly and looks like they are all dev dependencies? I was thinking we can do a release, then slowly take care of the dev deps since they dont actually affect the code at all.

aboimpinto:
The PR documentation was removed. 
I agree with you that this is a minor version bump. It's just upgrading the packages to the latest version or, in this case, removing the problematic package.

PR #976, opened 2026-06-27
feat(dapi): Create new neon-dapi package

raulduartep:
## Summary

- Add `@cityofzion/neon-dapi` — TypeScript implementation of the NEP-21 dAPI standard for Neo N3
- Provide `DapiProvider` interface, `DapiOperations` helper class, typed error system (`DapiError`/`DapiErrorCode`), and all NEP-21 types
- Include a browser-based conformance suite (`conformance/`) for validating wallet provider implementations against the spec

### DIFF
--- a/packages/neon-dapi/README.md
+++ b/packages/neon-dapi/README.md
@@ -38,15 +38,15 @@ Methods that require direct wallet UI interaction (`pickAddress`, `on`, `removeL
 
 Standardized error class thrown by all `DapiOperations` methods. Each error carries a `DapiErrorCode` so dApps can handle failure cases programmatically.
 
-### `ENetwork`
+### `DapiNetwork`
 
 Enum with the canonical Neo N3 network magic numbers.
 
 ```ts
-import { ENetwork } from "@cityofzion/neon-dapi";
+import { DapiNetwork } from "@cityofzion/neon-dapi";
 
-ENetwork.MAINNET // 860833102
-ENetwork.TESTNET // 894710606
+DapiNetwork.MAINNET // 860833102
+DapiNetwork.TESTNET // 894710606
 ```
 
 ## Conformance Suite
--- a/packages/neon-dapi/README.md
+++ b/packages/neon-dapi/README.md
@@ -1,0 +1,85 @@
+# neon-dapi
+
+## Overview
+
+TypeScript implementation of the [NEP-21 dAPI standard](https://github.com/neo-project/proposals/blob/master/nep-21.mediawiki) for Neo N3. Provides the `DapiProvider` interface that wallets must implement, along with `DapiOperations` — a helper class that handles the blockchain heavy lifting so wallet developers only need to wire up user interaction.
+
+## Installation
+
+```sh
+npm i @cityofzion/neon-dapi @cityofzion/neon-core
+```
+
+`@cityofzion/neon-core` is a peer dependency and must be installed alongside this package.
+
+## Concepts
+
+### Types
+
+All NEP-21 types are exported from this package and can be imported directly. Key types include:
+
+- **`DapiProvider`** — The interface wallets must implement. Defines the full NEP-21 surface exposed to dApps: account access, asset transfers, contract invocations, signing, and blockchain queries.
+- **`Account`**, **`Signer`**, **`Transaction`**, **`Block`** — Core blockchain data structures.
+- **`InvocationArguments`**, **`InvocationResult`** — Input and output for contract calls.
+- **`ContractParametersContext`** — Unsigned transaction context used in multi-sig flows (`makeTransaction` → `sign` → `relay`).
+- **`TransactionOptions`** — Fee and validity overrides for transaction building.
+- **`SignOptions`**, **`SignedMessage`** — Message signing configuration and result.
+- **`AuthenticationChallengePayload`**, **`AuthenticationResponsePayload`** — NEP-20 authentication payloads.
+- **`ProviderReadyEvent`**, **`ProviderRequestEvent`**, **`AccountChangedEvent`**, **`NetworkChangedEvent`** — Custom DOM events for the provider lifecycle and wallet state changes.
+- Primitive aliases — **`UInt160`**, **`UInt256`**, **`ECPoint`**, **`Integer`**, **`Base64Encoded`**, **`Network`** — string/number aliases with semantic names matching the NEP-21 spec.
+
+### `DapiOperations` class
+
+A helper that implements the blockchain and cryptographic operations required by `DapiProvider`. It is not a provider itself — it exists to reduce boilerplate by handling RPC calls, transaction building, signing, and fee calculation.
+
+Methods that require direct wallet UI interaction (`pickAddress`, `on`, `removeListener`) are intentionally omitted from `DapiOperations` because they depend on the specific wallet environment and cannot be generalized.
+
+### `DapiError`
+
+Standardized error class thrown by all `DapiOperations` methods. Each error carries a `DapiErrorCode` so dApps can handle failure cases programmatically.
+
+### `ENetwork`
+
+Enum with the canonical Neo N3 network magic numbers.
+
+```ts
+import { ENetwork } from "@cityofzion/neon-dapi";
+
+ENetwork.MAINNET // 860833102
+ENetwork.TESTNET // 894710606
+```
+
+## Conformance Suite
+
+The package ships a browser-based conformance suite (`conformance/`) that validates any injected NEP-21 provider against the full spec. Each method is exercised and results are reported as pass/fail directly in the page.
+
+The suite runs against the live injected provider — no mocking occurs. Useful for wallet developers to verify their `DapiProvider` implementation before release.
+
+### Running
+
+Serve the conformance directory with any static file server and open it in a browser that has a Neo wallet extension active:
+
+```sh
+# Using npx serve
+npx serve packages/neon-dapi/conformance
+```
+
+Then open `http://localhost:3000` (or whichever port your server uses) in the browser.
+
+> **Note:** Some tests require a connected account with funds on the target network.
+
+Click **Run All** to execute the full suite, or run individual tests from the UI.
+
+### `DapiErrorCode`
+
+| Code | Value | Meaning |
+|---|---|---|
+| `UNKNOWN` | 10000 | An unknown error has occurred. |
+| `UNSUPPORTED` | 10001 | The requested feature is not supported. |
+| `INVALID` | 10002 | The input data is in an invalid format. |
+| `NOTFOUND` | 10003 | The requested data doesn't exist. |
+| `FAILED` | 10004 | The contract execution failed. |
+| `TIMEOUT` | 10005 | The operation was cancelled due to timeout. |
+| `CANCELED` | 10006 | The operation was cancelled by the user. |
+| `INSUFFICIENT_FUNDS` | 10007 | The operation failed due to insufficient balance. |
+| `RPC_ERROR` | 10008 | An exception was thrown by the RPC server. |
--- a/packages/neon-dapi/__tests__/DapiOperations.ts
+++ b/packages/neon-dapi/__tests__/DapiOperations.ts
@@ -72,13 +72,14 @@ describe("DapiOperations.authenticate", () => {
     const payload = makePayload();
     const response = await ops.authenticate(payload);
 
-    const networkHex = u.num2hexstring(response.network, 4, true);
-    const nonceHex = u.num2hexstring(Number(payload.nonce), 8, true);
-    const timestampHex = u.num2hexstring(response.timestamp, 4, true);
-    const hashHex = account.scriptHash.replace(/^0x/i, "");
-    const actionHex = u.str2hexstring(payload.action);
-    const domainHex = u.str2hexstring(payload.domain);
-    const message = `${networkHex}${nonceHex}${timestampHex}${hashHex}${actionHex}${domainHex}`;
+    const bw = new u.BinaryWriter();
+    bw.writeUint64(Number(payload.nonce));
+    bw.writeUint32(response.timestamp);
+    bw.writeUint32(response.network);
+    bw.writeUInt160(account.scriptHash);
+    bw.writeVarString(payload.action);
+    bw.writeVarString(payload.domain);
+    const message = bw.toHex();
 
     const signatureHex = u.base642hex(response.signature);
     const isValid = wallet.verify(message, signatureHex, account.publicKey);
--- a/packages/neon-dapi/__tests__/DapiOperations.ts
+++ b/packages/neon-dapi/__tests__/DapiOperations.ts
@@ -1,0 +1,1293 @@
+import { u, wallet } from "@cityofzion/neon-core";
+
+import { DapiErrorCode } from "../src/DapiError";
+import { DapiOperations } from "../src/DapiOperations";
+import type { AuthenticationChallengePayload } from "../src/types";
+import { DapiNetwork } from "../src/constants";
+
+jest.mock("@cityofzion/neon-core", () => {
+  const actual = jest.requireActual("@cityofzion/neon-core");
+  return {
+    ...actual,
+    rpc: {
+      ...actual.rpc,
+      RPCClient: jest.fn().mockImplementation(() => ({})),
+    },
+  };
+});
+
+const ACCOUNT = "0x1234567890abcdef1234567890abcdef12345678";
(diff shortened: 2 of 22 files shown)


packages/neon-dapi/src/utils.ts
```diff
@@ -0,0 +1,236 @@
...
+      sc.ContractParam.fromJson({ type: arg.type, value: arg.value }),
+    );
+    scriptBuilder.emitContractCall({
+      operation: invocation.operation,
+      scriptHash: invocation.hash,
+      args,
+    });
+
+    if (invocation.abortOnFail) {
+      scriptBuilder.emit(sc.OpCode.ABORT);
```
ixje: This always aborts regardless of the actual result of the invocation. You want to use `ASSERT` instead. 

To be honest, I think this feature should not exist at all because it will only work as expected on invocations that return a boolean indicating success or not (e.g. nep11/17 transfer). Everything else will likely be experienced as unexpected behaviour. I would emit a warning to indicate the feature is not supported.

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+      return await this.invoke(
+        [
+          {
+            hash: asset,
+            operation: "transfer",
+            args: [
+              { type: "Hash160", value: from },
+              { type: "Hash160", value: to },
+              { type: "Integer", value: amount },
+              { type: "Any", value: data },
```
ixje: `data` is 
```
export type Argument = {
  type: ContractParameterType; // The type of the argument.
  value?: any; // The value of the argument. It can be omitted if the type is "Any".
};
```

that isn't going to work. Replace `{ type: "Any", value: data }` with `data`

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+        this.rpcClient,
+        [
+          {
+            hash: asset,
+            operation: "transfer",
+            args: [
+              { type: "Hash160", value: from },
+              { type: "Hash160", value: to },
+              { type: "Integer", value: amount },
+              { type: "Any", value: data },
```
ixje: https://github.com/CityOfZion/neon-js/pull/976/changes#r3491904193

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+        throw new DapiError(DapiErrorCode.INVALID, "Network is not supported");
+      }
+
+      const network = payload.networks[0];
+      const networkHex = u.num2hexstring(network, 1, true);
+      const nonceHex = u.num2hexstring(Number(payload.nonce), 8, true);
+      const timestampHex = u.num2hexstring(timestamp, 8, true);
+      const hashHex = this.account.scriptHash.replace(/^0x/i, "");
+
+      const hash = `${networkHex}${nonceHex}${timestampHex}${hashHex}`;
```
ixje: https://github.com/neo-project/proposals/blob/d23eadea66cf71817ede04a8b69beb4a59b55fd6/nep-20.mediawiki#response-data-to-sign seems to suggest there should be an `action` and `domain` as well. This raises the question; has this been tested against a real dapp that implements this?

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+      if (!payload.networks.length) {
+        throw new DapiError(DapiErrorCode.INVALID, "Networks is empty");
+      }
+
+      if (!payload.networks.includes(this.network)) {
+        throw new DapiError(DapiErrorCode.INVALID, "Network is not supported");
+      }
+
+      const network = payload.networks[0];
+      const networkHex = u.num2hexstring(network, 1, true);
```
ixje: network is supposed to be [4 bytes](https://github.com/neo-project/proposals/blob/d23eadea66cf71817ede04a8b69beb4a59b55fd6/nep-20.mediawiki?plain=1#L166), here it's limited to 1 byte

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+
+      if (!payload.networks.length) {
+        throw new DapiError(DapiErrorCode.INVALID, "Networks is empty");
+      }
+
+      if (!payload.networks.includes(this.network)) {
+        throw new DapiError(DapiErrorCode.INVALID, "Network is not supported");
+      }
+
+      const network = payload.networks[0];
```
ixje: I don't believe there is a guarantee that `payload.networks[0]` is the right network. 
```
const network = this.network
```

packages/neon-dapi/__tests__/DapiOperations.ts
```diff
@@ -0,0 +1,1293 @@
...
+
+    expect(response.timestamp).toBeGreaterThanOrEqual(before);
+    expect(response.timestamp).toBeLessThanOrEqual(after);
+  });
+
+  test("signature is cryptographically valid", async () => {
+    const payload = makePayload();
+    const response = await ops.authenticate(payload);
+
+    const networkHex = u.num2hexstring(response.network, 1, true);
```
ixje: https://github.com/CityOfZion/neon-js/pull/976/changes#r3491909712

packages/neon-dapi/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### `ENetwork`
+
+Enum with the canonical Neo N3 network magic numbers.
+
+```ts
+import { ENetwork } from "@cityofzion/neon-dapi";
+
+ENetwork.MAINNET // 860833102
+ENetwork.TESTNET // 894710606
```
ixje: `ENetwork` doesn't exist

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+   * @param txid - The transaction hash.
+   * @returns The application log containing all executions and notifications.
+   * @throws DapiError - INVALID | RPC_ERROR
+   */
+  async getApplicationLog(txid: UInt256): Promise<ApplicationLog> {
+    try {
+      const applicationLog = await this.rpcClient.getApplicationLog(txid);
+      return {
+        txid: applicationLog.txid,
+        executions: applicationLog.executions.map((execution) => ({
```
ixje, line 192: `exception` field is not mapped. I think that's important to include.

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+      if (options?.isBase64Encoded) {
+        hexMessage = u.base642hex(message);
+      } else {
+        hexMessage = u.str2hexstring(message);
+      }
+
+      const signature = wallet.sign(hexMessage, this.account.privateKey);
+
+      return {
+        payload: u.utf82base64(message),
```
ixje: can cause double base64 encoding if `message` was of type `Base64Encoded`

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -0,0 +1,839 @@
...
+    } catch (error) {
+      throw DapiError.parseError(error);
+    }
+  }
+
+  /**
+   * Returns the current connected account in the wallet.
+   * @returns Array containing the connected account.
+   */
+  getAccounts(): Account[] {
```
ixje, line 127: should this not return `Promise<Account[]>`?

https://github.com/neo-project/proposals/blob/462436528f9fe71993a1ffd0fa7df76cdb7f9641/nep-21.mediawiki?plain=1#L477

ixje (changes requested):
I think the general feedback is to test it against a real dapp that supports NEP-20/21. That would have captured at least half of the feedback

packages/neon-dapi/src/utils.ts
raulduartep: Since it's a simple feature to implement, I'll keep it to keep full compatibility with the dAPI interface.

packages/neon-dapi/src/DapiOperations.ts
raulduartep: You're right. I didn't notice that.

packages/neon-dapi/src/DapiOperations.ts
raulduartep: For most of the implementation, I used NeoLine as the source of truth since they already have dAPI support.

However, I also noticed that they are not fully compliant with the dAPI specification, as some methods don't return the expected responses. So it's possible that this is another case where their implementation doesn't match the specification.

Neoline code: https://github.com/NeoNEXT/neoline/blob/c02d086ac1ec5b7914af4123d612599c22358cce/src/app/popup/notification/neo3-authenticate/neo3-authenticate.component.ts#L118

I couldn't find any dApp that I could use to test all the dAPI methods, so I ended up using NeoLine's responses as a reference for my implementation. Because of that, I created a conformance "dApp" that attempts to verify the behavior of each dAPI method against the specification.

I'm going to update the implementation to also include the `action` and `domain` fields. Do you know if I should also encode these properties as hex before including them in the hash?

packages/neon-dapi/src/DapiOperations.ts
raulduartep: Yes, I forgot to change it.

packages/neon-dapi/src/DapiOperations.ts
raulduartep: I've also noticed that timestamp should be a 4 bytes, right ?

packages/neon-dapi/src/DapiOperations.ts
raulduartep: These methods are just utility functions to help the wallet comply with the dAPI specification. The wallet still needs to implement a lot of its own logic to make the dAPI integration work correctly. So, I can change them to return a `Promise` if you think it's more appropriate. It won't make any practical difference.

raulduartep:
@ixje Do you know a dApp that I can use to test this methods ?

packages/neon-dapi/src/DapiOperations.ts
ixje: yes

ixje:
> @ixje Do you know a dApp that I can use to test this methods ?

Sorry I don't. Probably best to ask on the NEO Discord

packages/neon-dapi/__tests__/DapiOperations.ts
```diff
@@ -72,11 +72,13 @@ describe("DapiOperations.authenticate", () => {
...
-    const networkHex = u.num2hexstring(response.network, 1, true);
+    const networkHex = u.num2hexstring(response.network, 4, true);
     const nonceHex = u.num2hexstring(Number(payload.nonce), 8, true);
-    const timestampHex = u.num2hexstring(response.timestamp, 8, true);
+    const timestampHex = u.num2hexstring(response.timestamp, 4, true);
     const hashHex = account.scriptHash.replace(/^0x/i, "");
-    const message = `${networkHex}${nonceHex}${timestampHex}${hashHex}`;
+    const actionHex = u.str2hexstring(payload.action);
+    const domainHex = u.str2hexstring(payload.domain);
+    const message = `${networkHex}${nonceHex}${timestampHex}${hashHex}${actionHex}${domainHex}`;
```
ixje: I don't think this is correct. Note in the spec https://github.com/neo-project/proposals/blob/d23eadea66cf71817ede04a8b69beb4a59b55fd6/nep-20.mediawiki#response-data-to-sign for the `type` column is says `var_str`. Now of course they didn't document what this means, but I'm guessing it's the same as https://github.com/CityOfZion/neon-js/blob/9ec56857a70d4db4e7a02e5eff3509609df0928e/packages/neon-core/src/sc/ScriptBuilder.ts#L153

where the length in encoded in a specific format and prepended to the string itself

packages/neon-dapi/src/DapiOperations.ts
```diff
@@ -93,20 +94,22 @@ export class DapiOperations {
...
 
-      const network = payload.networks[0];
-      const networkHex = u.num2hexstring(network, 1, true);
+      const networkHex = u.num2hexstring(this.network, 4, true);
       const nonceHex = u.num2hexstring(Number(payload.nonce), 8, true);
-      const timestampHex = u.num2hexstring(timestamp, 8, true);
+      const timestampHex = u.num2hexstring(timestamp, 4, true);
       const hashHex = this.account.scriptHash.replace(/^0x/i, "");
+      const actionHex = u.str2hexstring(payload.action);
+      const domainHex = u.str2hexstring(payload.domain);
```
ixje: I commented in the wrong file but the same comment holds; https://github.com/CityOfZion/neon-js/pull/976/changes/f8321a6f6ed95ddc9d0664b6fce6aced4ecc4782..0a61dde58fda727b5ccf929cce0c0e11b2ed26d4#r3513194118

ixje (changes requested):

packages/neon-dapi/__tests__/DapiOperations.ts
raulduartep: Yes, I noticed that. However, I was looking at the `sign` method in `neon-js`, and it expects to receive a hex string. Because of that, I thought using these properties as raw strings could introduce some inconsistency.
Do you think I should include these properties as raw strings instead?

packages/neon-dapi/src/DapiOperations.ts
raulduartep: https://github.com/CityOfZion/neon-js/pull/976#discussion_r3514179020

packages/neon-dapi/__tests__/DapiOperations.ts
ixje: I think that should be fine. 

Here's what I think Erik Zhang expects;

```ts
        let bw = new BinaryWriter()
        const nonce = 1234
        const timestamp = 1783006981
        const network = 860833102
        const account = UInt160.fromString("0x30e9d4a4758827361c3b51a0e8460b067e58b1db")
        bw.writeUint64(nonce)
        bw.writeUint32(timestamp)
        bw.writeUint32(network)
        bw.writeSerializable(account)
        bw.writeVarString("some_action")
        bw.writeVarString("domain")
        console.log(uint8ToHex(bw.toArray()))
```
that will print
```
d2040000000000000587466a4e454f33dbb1587e060b46e8a0513b1c36278875a4d4e9300b736f6d655f616374696f6e06646f6d61696e
```

This is the `BinaryWriter` class for reference

```ts
const MAX_SCRIPT_SIZE = 0xFFFF;

export class BinaryWriter {
    // TODO: make auto growable with resize() but can't due to ArrayBuffer options
    // not being available. See https://github.com/microsoft/TypeScript/pull/58573
    // fixed size now
    private _stream = new Uint8Array(MAX_SCRIPT_SIZE);
    private _view = new DataView(this._stream.buffer);
    private _position = 0;

    writeUint32(value: number): void {
        this._view.setUint32(this._position, value, true);
        this._position += 4;
    }

    writeUint64(value: number | bigint): void {
        let v = value;
        if (typeof value === "number") {
            v = BigInt(value);
        }
        this._view.setBigUint64(this._position, v as bigint, true);
        this._position += 8;
    }

    writeVarInt(value: number): void {
        if (value < 0) {
            throw new Error("value too small");
        } else if (value < 0xFD) {
            this.writeBytes(new Uint8Array([value]));
        } else if (value <= 0xFFFF) {
            this.writeBytes(new Uint8Array([0xFD]));
            this.writeUint16(value);
        } else if (value <= 0xFFFFFFFF) {
            this.writeBytes(new Uint8Array([0xFE]));
            this.writeUint32(value);
        } else {
            this.writeBytes(new Uint8Array([0xFF]));
            this.writeUint64(value);
        }
    }

    writeVarString(value: string): void {
        const data = new TextEncoder().encode(value);
        this.writeVarInt(data.length);
        this._stream.set(data, this._position);
        this._position += data.length;
    }

    writeSerializable(obj: Serializable): void {
        obj.serialize(this);
    }

    toArray(): Uint8Array {
        return this._stream.slice(0, this._position);
    }

    get length(): number {
        return this._position;
    }
}
```
Note; this is not in `neon-js` (it comes from a personal  local project) but it should allow you to follow what is happening if the output doesn't match. For the `UInt160` specifically be aware that it strips the `0x` prefix and reverses the bytes. All values are randomly chosen, but I'm 98% confident this is what Erik Zhang intended.

ixje:
Let's go for it. We will never know 100% sure if it matches the spec as intended until they provide a reference implementation or dApp.

ixje (approved):

